Issue #20, opened 2024-10-02
Custom Gitlab URL not reflected in generated .netrc


tomusher:
When specifying a custom `GitLabURL` in the config, this URL is not reflected in the `.netrc` file generated by `entrypoint.sh`, so auth does not work without overriding the generated file.

zapp42:
Perfect timing. You saved me some debugging. Thank you.
Confirmed.
Symptoms: The group hierarchy is created in the .sourcebot/repos folder, but no actual repository is cloned.

brendan-kellam:
Hey thanks for raising. I will take a sec and get a custom GitLab instance going and will then take a look

PR #21, opened 2024-10-02
Use GitLabURL in .netrc

zapp42:
Quick hack to fix issue #20


brendan-kellam:
hey @zapp42, thanks for the PR! I think for now I would rather keep the hostname explicitly defined via a environment variable rather than pulling it from the config. It's simpler, but the tradeoff to the user is that you need to define the url and hostname in two places (i.e., in `GitLabUrl` and the env var).

I'm thinking allot of this stuff will be changing anyways (e.g., to enable #27), so I think simpler is better for now.

See #26

PR #26, opened 2024-10-03
Add env vars for overriding GitHub / GitLab hostnames in the `.netrc` file.

### LINKED ISSUE #20: Custom Gitlab URL not reflected in generated .netrc

brendan-kellam:
Adds override env vars `GITLAB_HOSTNAME` and `GITHUB_HOSTNAME` to support authenticating against self-hosted GitLab / GitHub instances.

Fixes #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `GITLAB_HOSTNAME` and `GITHUB_HOSTNAME` environment variables to allow overriding the default hostnames for GitLab and GitHub.
+
 ## [1.0.0] - 2024-10-01
 
 ### Added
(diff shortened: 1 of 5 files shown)


msukkari (approved):

loris:
@brendan-kellam Thank you for fix, however why do you use the env vars to build the .netrc file, instead of the settings from the config file? I'm asking because i'm trying to index repos from multiple self-hosted gitlab servers (so I have more than gitlab one hostname and token). It seems limited to just 1 if we rely on env var, while the config file is more flexible since you can have any number of entries in the Configs array, right?

brendan-kellam:
Hey @loris - yea that's correct. Using env vars was a quick fix for the single self-hosted server scenario.

We are working on overhauling the config format atm and plan to support multiple host authentication (via the config) at that time. We expect to have that in sometime next week. See: #27

PR #38, opened 2024-10-15
fix: fix  javascript btoa function can not handle Unicode strings issue

ttys3:
image

see https://web.dev/articles/base64-encoding#strings_in_unicode_and_javascript

https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings

https://stackoverflow.com/questions/8936984/uint8array-to-string-in-javascript


brendan-kellam (approved):

brendan-kellam:
Thanks for the fix!!

PR #42, opened 2024-10-16
Config format V2

### LINKED ISSUE #27: Support authenticating against multiple GitLab / GitHub hosts
Currently, we are limited to authenticating against a single GitLab / GitHub host (i.e., you can authenticate against `gitlab.com` or `gitlab.example.com`, but not both).

Ideally, we would have support for indexing across multiple hosts.

### LINKED ISSUE #35: Add logging / UI cues when a repository fails to clone
I seem to have hit a problem where Sourcebot only indexes repositories with `public` visibility and not `internal` or `private` visibility on our self-hosted GitLab instance.

I have create a dedicated Sourcebot account in GitLab (marked as an external user) and generated an "Impersonation Token" (essentially the same as a "Personal Access Token) for the account via the GitLab Admin interface.

The Sourcebot account has been added as a member to a few different repositories and groups (of which most repositories have their visibility set to either "Internal" or "Private").

The "Impersonation Token" has been set in the `GITLAB_TOKEN` environment variable when starting the container. I can see requests to the GitLab API coming from Sourcebot and can confirm that the responses for the `/api/v4/projects?membership=true&order_by=id&per_page=100&sort=asc` endpoint contain all the repositories that the Sourcebot account has access to are indeed being returned. However, it seems that Sourcebot is only indexing the repositories which have a visibility of `public`.

I'm not sure if this is indented behaviour, or a bug, but it would be good if there were a way ensure that Sourcebot indexes GitLab repositories with a visibility of `internal` or `private`.

brendan-kellam:
This PR introduces a new configuration schema that is more expressive.

Example config: 
```json
{
    "repos": [
        {
            "type": "github",
            "token": "secret-token",
            "orgs": [
                "sourcebot-dev",
                "commaai"
            ],
            "users": [
                "DHH"
            ],
            "repos": [
                "torvalds/linux",
                "my-org/my-repo"
            ],
            "exclude": {
                "repos": [
                    "sourcebot-dev/sourcebot"
                ],
                "archived": true,
                "forks": true
            }
        },
        {
            "type": "gitlab",
            "token": "secret-token",
            "groups": [
                "my-group"
            ],
            "users": [
                "user1",
                "user2"
            ],
            "projects": [
                "my-group/project1",
                "my-group/my-sub-group/project2"
            ],
            "exclude": {
                "projects": [
                    "my-group/my-project3"
                ],
                "archived": true,
                "forks": true
            }
        }
    ]
}
```

With the new configuration schema comes a new index server, located in `packages/backend`. This has the same functionality as `zoekt-indexserver` (it will clone repos and periodically re-index them), but with improved logging s.t., it's now easier to determine what the index server is doing.

Fixes #35 
Fixes #27


packages/backend/package.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "@sourcebot/backend",
+    "version": "0.1.0",
+    "private": true,
+    "main": "index.js",
+    "type": "module",
+    "scripts": {
+        "dev:watch": "yarn generate:types && tsc-watch --onSuccess \"yarn dev --configPath ../../config.json --cacheDir ../../.sourcebot\"",
```
brendan-kellam, line 8: We probably want to double check that these arguments for `--configPath` and `--cacheDir` work when calling `dev:watch` from the workspace root.

package.json
```diff
@@ -1,83 +1,15 @@
...
-    "postcss": "^8",
-    "tailwindcss": "^3.4.1",
-    "typescript": "^5"
-  }
+    "private": true,
+    "workspaces": [
+        "packages/*"
+    ],
+    "scripts": {
+        "dev": "npm-run-all --print-label --parallel web zoekt:webserver zoekt:indexserver",
```
brendan-kellam, line 7: This needs to be updated

README.md
```diff
@@ -70,23 +70,25 @@ Sourcebot supports indexing and searching through public and private repositorie
...
+                "repos": [
+                    "ggerganov/llama.cpp"
+                ]
             }
         ]
     }' > my_config.json
     ```
 
-    (For more examples, see [example-config.json](./example-config.json). For additional usage information, see the [configuration schema](./schemas/index.json)).
+    >[!NOTE]
```
msukkari, line 90: nit: these NOTE highlights seem to be broken?

msukkari (dismissed):
🔥🔥🔥 tested working on my side (public/private github and gitlab, as well as self-hosted gitlab)

README.md
brendan-kellam: ah kk - I think GitHub doesn't like it when you indent these. Will change.

PR #45, opened 2024-10-20
Gitea support

### LINKED ISSUE #13: FR: Support indexing code hosted on a local Gitea instance
[Gitea](https://github.com/go-gitea/gitea) is a popular Git server in the self hosting community. I would love to be able to pair it with a self hosted sourcebot instance. Thanks!

brendan-kellam:
This PR adds support for Gitea to Sourcebot.

Fixes #13


msukkari (approved):
tested with public and private org/user/repo

PR #47, opened 2024-10-24
Add discord button to nav bar and version in settings drop down

msukkari:


.gitignore
```diff
@@ -163,4 +163,5 @@ dist
 .sourcebot
 /bin
 /config.json
-.DS_Store
\ No newline at end of file
+.DS_Store
+.vscode
```
brendan-kellam, line 167: I think we want to keep the .vscode directory - the settings.json i have in there makes it s.t., config files default to "Json with comments" instead of just "Json" (since our config schema supports comments).

brendan-kellam (dismissed):

brendan-kellam (approved):

PR #56, opened 2024-11-01
Local directory support

### LINKED ISSUE #19: FR: Support indexing of local directories
Hej :)

It would be great if Sourcebot could handle local directory with one or more git repositories.

(Also, any pointers can be useful to implement such support by myself - also great learning) :) 

Thanks! :)

brendan-kellam:
This PR expands the configuration schema to support local directories. For example:
```jsonc
{
    "repos": [
        {
            "type": "local",
            "path": "/path/to/my/repo",
            "watch": true,
            "exclude": {
                "paths": [
                    ".git",
                    "node_modules"
                ]
            }
        }
    ]
}
```

Fixes #19

### DIFF
--- a/packages/backend/src/zoekt.ts
+++ b/packages/backend/src/zoekt.ts
@@ -1,6 +1,8 @@
 import { exec } from "child_process";
 import { AppContext, GitRepository, LocalRepository } from "./types.js";
 
+const ALWAYS_EXCLUDED_DIRS = ['.git', '.hg', '.svn'];
+
 export const indexGitRepository = async (repo: GitRepository, ctx: AppContext) => {
     return new Promise<{ stdout: string, stderr: string }>((resolve, reject) => {
         exec(`zoekt-git-index -index ${ctx.indexPath} ${repo.path}`, (error, stdout, stderr) => {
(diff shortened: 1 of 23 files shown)


packages/backend/src/zoekt.ts
```diff
@@ -0,0 +1,35 @@
...
+            resolve({
+                stdout,
+                stderr
+            });
+        })
+    });
+}
+
+export const indexLocalRepository = async (repo: LocalRepository, ctx: AppContext, signal?: AbortSignal) => {
+    const excludedDirs = repo.excludedPaths.length > 0 ? repo.excludedPaths : ['.git', '.hg', '.svn'];
```
msukkari, line 20: nit: do we always want to ignore .git, .hg, and .svn? If so this may cause these dirs to not be ignored if they provide exclude paths and forget to include those

msukkari (dismissed):

packages/backend/src/zoekt.ts
brendan-kellam: Yea this makes sense, I think it's a fair assumption that these will always be excluded.

PR #58, opened 2024-11-06
Multi branch / tag support

### LINKED ISSUE #29: FR: Support indexing of branches
Sometimes I keep variants of my code on branches without intending to merge it into the default branch. Would be cool to see those in Sourcebot too.

brendan-kellam:
The following PR adds support for indexing multiple revisions (branches & tags) of a given GitHub, GitLab, or Gitea repository. Configuration of which revisions are included is made via the `revisions` property:
```jsonc
{
    "repos": [
        {
            "type": "github",
            "revisions": {
                 // Specify branches to index...
                "branches": [
                    "main",
                    "release/*"
                ],
                // ... or specify tags
                "tags": [
                    "v*.*.*"
                ]
            },
            "repos": [
                "org/repoa",
                "org/repob"
            ]
        }
   ]
}
```

The branch / tag values support glob patterns. In this case, we will match against the `main` branch + all branches starting with `release/`, as well as all tags that match the semver `v*.*.*` format. The branches / tags in a given repo (in this case, `org/repoa` and `org/repob`) **and** that match a glob pattern will be included during indexing. The `revisions` object is similarly supported for GitLab and GItea.

To search a particular revision, the `revision:` or `rev:` qualifier can be included in a query. For example:
image

The original zoekt `branch:` or `b:` qualifier may also be used.

Fixes #29


msukkari:
nit: scroll bars in repos page seem broken on windows
image

msukkari:
what does the "+ 2" here mean?
image

msukkari:
what does the "+ 2" here mean? If it's supposed to reference how many branch hits there are, would it be better to instead mention this in the search results at the top (ex. Found X matches in Y files across Z branches)? It's not really intuitive what this number means at first glance
image

msukkari (dismissed):

brendan-kellam:
> what does the "+ 2" here mean? If it's supposed to reference how many branch hits there are, would it be better to instead mention this in the search results at the top (ex. Found X matches in Y files across Z branches)? It's not really intuitive what this number means at first glance image

We could surface a "Found X matches in Y files across Z branches" at the top, but we still probably need a per-match UI since different matches may exist in a different number of branches.

PR #63, opened 2024-11-09
Add additional telemetry

brendan-kellam:
This PR adds two additional PostHog events:
- `upgrade` - fired when a installation's version is changed
- `repo_created` - fired when a repository is first ingested, prior to indexing.
- `repo_synced` - fired whenever a repository is indexed.


Dockerfile
```diff
@@ -55,13 +55,14 @@ RUN echo "Sourcebot Version: $SOURCEBOT_VERSION"
 ENV SOURCEBOT_LOG_LEVEL=info
 
 # @note: This is also set in .env
-ENV NEXT_PUBLIC_POSTHOG_KEY=phc_VFn4CkEGHRdlVyOOw8mfkoj1DKVoG6y1007EClvzAnS
+ENV POSTHOG_KEY=phc_VFn4CkEGHRdlVyOOw8mfkoj1DKVoG6y1007EClvzAnS
+ENV NEXT_PUBLIC_POSTHOG_KEY=$POSTHOG_KEY
```
msukkari, line 59: out of curiosity whats the purpose of adding the `POSTHOG_KEY` env var?

msukkari (approved):

Dockerfile
brendan-kellam: Because the backend uses the `POSTHOG_KEY` env var and not `NEXT_PUBLIC_POSTHOG_KEY`. It's not strictly needed since `POSTHOG_KEY` is also defined in the `.env` file. This is the same for web (`NEXT_PUBLIC_POSTHOG_KEY` is defined in it's `.env`), so I figured we should be consistent

PR #76, opened 2024-11-18
WIP: Add basic gerrit support

KonradStanski:
This commit adds support for gerrit repos.
It will either fetch all repos, or the specified ones. Branch fetching is supported
Tags are not yet supported
Auth is not yet supported (not necessary for default gerrit setup) Need to add useage of gerritkit rest api npm module

DO NOT MERGE YET
PLEASE SQUASH ON MERGE

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9950 @@
+{
+  "name": "sourcebot",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "workspaces": [
+        "packages/*"
+      ],
+      "devDependencies": {
+        "npm-run-all": "^4.1.5"
+      }
+    },
+    "node_modules/@alloc/quick-lru": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz",
+      "integrity": "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@apidevtools/json-schema-ref-parser": {
+      "version": "11.7.0",
+      "resolved": "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-11.7.0.tgz",
+      "integrity": "sha512-pRrmXMCwnmrkS3MLgAIW5dXRzeTv6GLjkjb4HmxNnvAKXN1Nfzp4KmGADBQvlVUcqi+a5D+hfGDLLnd5NnYxog==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@jsdevtools/ono": "^7.1.3",
+        "@types/json-schema": "^7.0.15",
+        "js-yaml": "^4.1.0"
+      },
+      "engines": {
+        "node": ">= 16"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/philsturgeon"
+      }
+    },
+    "node_modules/@babel/runtime": {
+      "version": "7.25.7",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.25.7.tgz",
+      "integrity": "sha512-FjoyLe754PMiYsFaN5C94ttGiOmBNYTf6pLr4xXHAT5uctHb092PBszndLDR5XA/jghQvn4n7JMHl7dmTgbm9w==",
+      "license": "MIT",
+      "dependencies": {
+        "regenerator-runtime": "^0.14.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@codemirror/autocomplete": {
+      "version": "6.18.1",
+      "resolved": "https://registry.npmjs.org/@codemirror/autocomplete/-/autocomplete-6.18.1.tgz",
+      "integrity": "sha512-iWHdj/B1ethnHRTwZj+C1obmmuCzquH29EbcKr0qIjA9NfDeBDJ7vs+WOHsFeLeflE4o+dHfYndJloMKHUkWUA==",
+      "license": "MIT",
+      "dependencies": {
+        "@codemirror/language": "^6.0.0",
+        "@codemirror/state": "^6.0.0",
+        "@codemirror/view": "^6.17.0",
+        "@lezer/common": "^1.0.0"
+      },
+      "peerDependencies": {
+        "@codemirror/language": "^6.0.0",
+        "@codemirror/state": "^6.0.0",
+        "@codemirror/view": "^6.0.0",
+        "@lezer/common": "^1.0.0"
+      }
+    },
+    "node_modules/@codemirror/commands": {
+      "version": "6.6.2",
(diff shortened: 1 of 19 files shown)


KonradStanski:
Should we add the yarn.lock file to gitignore?

KonradStanski:
This works as is with public gerrit instance. Try:
```
{
    "$schema": "https://raw.githubusercontent.com/sourcebot-dev/sourcebot/main/schemas/v2/index.json",
    "repos": [
        {
            "type": "gerrit",
            "url": "https://review.opendev.org",
            "limit": 5,
            "exclude": {
                "projects": [
                    "All-Users",
                    "All-Projects"
                ]
            },
            "revisions": {
                "branches": [
                    "refs/heads/v2.0",
                    "refs/heads/v2.1"
                ],
                "tags": [
                    "applier-v2.1.0"
                ]
            }
        }
    ]
}

```
Ignore tags, it's not implemented yet.

brendan-kellam:
> Should we add the yarn.lock file to gitignore?

No we want `yarn.lock` to be checked into source control. See [here](https://classic.yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control).

schemas/v2/index.json
```diff
@@ -399,6 +405,66 @@
...
+                    "examples": [
+                        "https://gerrit.example.com"
+                    ],
+                    "pattern": "^https?:\\/\\/[^\\s/$.?#].[^\\s]*$"
+                },
+                "limit":{
+                    "type": "integer",
+                    "description": "The maximum number of projects to fetch."
+                },
+                "projects": {
```
brendan-kellam, line 428: I'm unfamiliar with Gerrit - does it have a concept of a "organization" or similar that contains a set of projects? If so, it would be good to include that in the schema s.t., we can allow indexing every repo in a organization.

packages/backend/src/gerrit.ts
```diff
@@ -0,0 +1,179 @@
...
+      fetchProjects(url, config)
+   );
+
+   logger.debug(`Fetched ${Object.keys(projects).length} projects in ${durationMs}ms.`);
+
+   let projectNames = Object.keys(projects);
+
+   // If specific repos are specified in config, filter the projects
+   if (config.projects && config.projects.length > 0) {
+      projectNames = projectNames.filter((name) => config.projects!.includes(name));
```
brendan-kellam, line 68: nit: instead of getting all projects and then filtering that list down, could we instead specifically fetch each of the `project`s individually using [this api](https://gerrit-review.googlesource.com/Documentation/rest-api-projects.html#:~:text=Projects%20REST%20endpoint.-,Get%20Project,-%27GET%20/projects/)?

packages/backend/src/gerrit.ts
```diff
@@ -0,0 +1,179 @@
...
+   }
+
+   // If specific repos are excluded in config, filter the projects
+   logger.debug(`Excluding repos: ${config.exclude?.projects}`);
+   if (config.exclude && config.exclude.projects && config.exclude.projects.length > 0) {
```
brendan-kellam, line 73: nit: you should consider using `excludeReposByName` in `utils.ts`, which the other code hosts are using (which lets us keep our exclusion rules consistent :))

brendan-kellam (changes requested):
Thanks for working on this! I left a few comments and thoughts that we should consider

schemas/v2/index.json
```diff
@@ -399,6 +405,66 @@
...
+                "url": {
+                    "type": "string",
+                    "format": "url",
+                    "description": "The URL of the Gerrit host.",
+                    "examples": [
+                        "https://gerrit.example.com"
+                    ],
+                    "pattern": "^https?:\\/\\/[^\\s/$.?#].[^\\s]*$"
+                },
+                "limit":{
```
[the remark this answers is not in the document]
KonradStanski: Testing purposes mostly atm. I'm using https://review.opendev.org/ for testing since I don't have a publicly available gerrit instance for testing. I don't want to clone their whole org as this might raise some scrutiny.

packages/backend/src/gerrit.ts
```diff
@@ -0,0 +1,179 @@
...
+
+   const text = await response.text();
+
+   // Gerrit prepends ")]}'\n" to prevent XSSI attacks; remove it
+   const jsonText = text.replace(")]}'\n", '');
+   const data = JSON.parse(jsonText);
+   return data;
+}
+
+// TODO: Implement fetching tags from Gerrit
```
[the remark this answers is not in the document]
KonradStanski: ack

package-lock.json
```diff
@@ -0,0 +1,9950 @@
+{
```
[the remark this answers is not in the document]
KonradStanski: fat fingered `npm i` by instinct, thanks for the catch. This needs lots of cleanup atm.

packages/backend/src/gerrit.ts
```diff
@@ -0,0 +1,179 @@
...
+   let response = null;
+   response = await fetch(projectsEndpoint);
+
+   if (!response.ok) {
+      throw new Error(`Failed to fetch projects from Gerrit: ${response.statusText}`);
+   }
+
+   const text = await response.text();
+
+   // Gerrit prepends ")]}'\n" to prevent XSSI attacks; remove it
```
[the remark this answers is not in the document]
KonradStanski: https://gerrit-review.googlesource.com/Documentation/rest-api.html#:~:text=To%20prevent%20against%20Cross%20Site%20Script%20Inclusion%20(XSSI)%20attacks%2C%20the%20JSON%20response%20body%20starts%20with%20a%20magic%20prefix%20line%20that%20must%20be%20stripped%20before%20feeding%20the%20rest%20of%20the%20response%20body%20to%20a%20JSON%20parser%3A

Will do. It's on their main docs page.

schemas/v2/index.json
KonradStanski: Gerrit does not have orgs. Project organization is done via path naming conventions.
IE projects belonging to "org" airship are all under the airship/ path for the project name.
To achieve org import, I think all we need is support for globbing for the projects json. Or do you think there is some more appropriate mechanism?

The other mechanism here is member "groups" which is used as an ACL mechanism wrt projects. You could in theory specify member groups and then query the /access?project={} endpoint to determine for each project if that groups has access...

I'm not sure if this is really the way most people would want to use sourcebot? I'm not a gerrit expert.
I intuit that either indexing everything available, or globbing the repo name should be sufficient.

packages/backend/src/gerrit.ts
KonradStanski: ack

schemas/v2/index.json
brendan-kellam: Spoke offline - I think glob patterns make the most sense here since hierarchy is encoded within a path. To index everything in the `airship` path, it feels intuative to do:
```jsonc
{
    "repos": [
        {
            "type": "gerrit",
            "projects": [
                "airship/**"
            ]
        }
    ]
}
```

Note: glob patterns should probably also be supported for the `exclude` object to, allowing you to exclude sub-paths:

```jsonc
{
    "repos": [
        {
            "type": "gerrit",
            "repos": [
                "airship/**"
            ],
           "exclude": {
                "projects": [
                   "airship/subfolder/**"
                ] 
           }
        }
    ]
}
```

schemas/v2/index.json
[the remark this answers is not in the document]
brendan-kellam: Makes sense - once we have glob pattern support, this will be less of a issue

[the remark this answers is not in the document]
brendan-kellam: Also that's useful info about https://review.opendev.org/ - I now see what you mean about paths being used to organize projects.

packages/backend/src/gerrit.ts
brendan-kellam: If we do glob patterns, fetching all projects makes sense

PR #85, opened 2024-11-21
Search suggestions

brendan-kellam:
The following PR introduces search suggestions:

https://github.com/user-attachments/assets/260acc4f-0a97-464a-b26b-284af4ca1603

The motivation for building search suggestions is two-fold: (1) to make the zoekt query language more approachable by presenting all available options to the user, and (2) make it easier for power-users to craft complex queries.

The meat-n-potatoes of this change are concentrated in `searchBar.tsx` and `searchSuggestionBox.tsx`. The suggestions box works by maintaining a state-machine of "modes". By default, the box is in the `refine` mode, where suggestions for different prefixes (e.g., `repo:`, `lang:`, etc.) are suggested to the user. When one of these prefixes is matched, the state-machine transitions to the corresponding mode (e.g., `repository`, `language`, etc.) and surfaces suggestions for that mode (if any).

The query is split up into parts by spaces " " (e.g., `'test repo:hello'` -> `['test', 'repo:hello']`). See `splitQuery`. The part that has the cursor over it is considered the active part. We evaluate which mode the state machine is in based on the active part. When a suggestion is clicked, we only modify the active part of the query.

Three modes are currently missing suggestion data: `file` (file names), `revision` (branch / tag names), and `symbol` (symbol names). In future PRs, we will need to introduce endpoints into the backend to allow the frontend to fetch this data and surface it as suggestions..


packages/web/src/app/components/searchBar/searchSuggestionsBox.tsx
```diff
@@ -0,0 +1,423 @@
...
+                </div>
+            )}
+        </div>
+    )
+});
+
+SearchSuggestionsBox.displayName = "SearchSuggestionsBox";
+export { SearchSuggestionsBox };
+
+const splitQuery = (query: string, seperator: string, cursorPos: number) => {
```
brendan-kellam, line 385: I want to add some UTs for this function since it has a few edge-cases

msukkari (approved):

PR #105, opened 2024-12-02
Filter results css

KonradStanski:
This fixes a couple things which seemed to not be working with the filter results panel:
1. It makes the filter result panel 100% of parent height, and set the repo filtering to be at max 50%. It then lets the language filter fill the remaining height.

2. This fixes the repo display name text such that it does not split with long repo names onto multiple lines, instead showing an ellipsis and gracefully being cut off.
It also puts the count into a box.

3. It also adds cutoff for both search bars to display the number of languages/ repos matched, and cut them off at 100.
In this case the text becomes 100+

4. Do the same for the result counts but cut them off at 999. It makes the text 999+
Before:
Screenshot 2024-12-02 at 8 14 33 PM

After:
Screenshot 2024-12-02 at 8 15 40 PM

Before:
Screenshot 2024-12-02 at 8 17 51 PM

After:
Screenshot 2024-12-02 at 8 18 17 PM

### DIFF
--- a/packages/web/src/app/search/components/filterPanel/entry.tsx
+++ b/packages/web/src/app/search/components/filterPanel/entry.tsx
@@ -25,10 +25,14 @@ export const Entry = ({
     },
     onClicked,
 }: EntryProps) => {
-
+    let countText = count.toString();
+    if (count > 999) {
+        countText = "999+";
+    }
     return (
         <div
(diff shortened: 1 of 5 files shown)


brendan-kellam:
Thank you! Could you paste in a before / after screenshot in your PR description?

KonradStanski:
> Thank you! Could you paste in a before / after screenshot in your PR description?

Just did this.

brendan-kellam (approved):

PR #106, opened 2024-12-03
Repository table CSS tweaks

KonradStanski:
This adds a few CSS tweaks to the repository search table.

1.
We add a `table-fixed` property and a column width for the name column. This prevents the table layout from jumping around when changing pages which have different max repo name lengths.

2.
Added a pagination page counter, and moved the previous and next buttons to the top right.

3.
Get the table to be responsive and grow to fill the available screen space

The changes are mostly dynamic, so don't really show in before and after pictures, so try this out and change the page size and shape. Should work for various zoom levels.

Before:
Screenshot 2024-12-02 at 9 52 11 PM

Screenshot 2024-12-02 at 9 52 21 PM

After:
Screenshot 2024-12-02 at 9 52 49 PM

### DIFF
--- a/packages/web/src/app/repos/columns.tsx
+++ b/packages/web/src/app/repos/columns.tsx
@@ -25,7 +25,7 @@ export const columns: ColumnDef<RepositoryColumnInfo>[] = [
     {
         accessorKey: "name",
         header: "Name",
-        size: 300,
+        size: 250,
         cell: ({ row }) => {
             const repo = row.original;
             const url = repo.url;
@@ -34,7 +34,7 @@ export const columns: ColumnDef<RepositoryColumnInfo>[] = [
             return (
                 <div className="flex flex-row items-center gap-2">
                     <span
-                        className={!isRemoteRepo ? "cursor-pointer text-blue-500 hover:underline": ""}
+                        className={"whitespace-normal break-all " + (!isRemoteRepo ? "cursor-pointer text-blue-500 hover:underline" : "")}
                         onClick={() => {
                             if (!isRemoteRepo) {
                                 window.open(url, "_blank");
@@ -62,7 +62,8 @@ export const columns: ColumnDef<RepositoryColumnInfo>[] = [
                     {branches.map(({ name, version }, index) => {
                         const shortVersion = version.substring(0, 8);
                         return (
-                            <span key={index}>
+                            <span
+                                key={index}>
                                 {name}
                                 @
                                 <span
--- a/packages/web/src/app/repos/columns.tsx
+++ b/packages/web/src/app/repos/columns.tsx
@@ -25,6 +25,7 @@ export const columns: ColumnDef<RepositoryColumnInfo>[] = [
     {
         accessorKey: "name",
         header: "Name",
+        size: 300,
         cell: ({ row }) => {
             const repo = row.original;
             const url = repo.url;
--- a/packages/web/src/components/ui/data-table.tsx
+++ b/packages/web/src/components/ui/data-table.tsx
@@ -40,7 +40,7 @@ export function DataTable<TData, TValue>({
     const [sorting, setSorting] = React.useState<SortingState>([])
     const [columnFilters, setColumnFilters] = React.useState<ColumnFiltersState>(
         []
-      )
+    )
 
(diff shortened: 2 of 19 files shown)


KonradStanski:
Thinking about this, I should have maybe split this into 2 diff pr's. The first part is making the table jump around less when changing pages and moving the buttons up top.
The second change of making the table dynamically change length based on window size is... more complicated, and perhaps not the perfect solution. If it's not what we want, I can go and remove it.

brendan-kellam:
> Thinking about this, I should have maybe split this into 2 diff pr's. The first part is making the table jump around less when changing pages and moving the buttons up top. The second change of making the table dynamically change length based on window size is... more complicated, and perhaps not the perfect solution. If it's not what we want, I can go and remove it.

Up to you - I'm also cool with this going in as one PR.

There might be a simpler solution where we can make the table itself scrollable and then assign it a fixed height. This might be more flexible for other scenarios in the future when we aren't dealing with a table that is expected to fill the vertical space of the screen.. but no strong preference one way or the other.

KonradStanski:
should be good to go, just maybe squash merge since I messed up the git history pretty bad.

packages/web/src/app/repos/columns.tsx
```diff
@@ -33,7 +34,7 @@ export const columns: ColumnDef<RepositoryColumnInfo>[] = [
             return (
                 <div className="flex flex-row items-center gap-2">
                     <span
-                        className={!isRemoteRepo ? "cursor-pointer text-blue-500 hover:underline": ""}
+                        className={"whitespace-normal break-all " + (!isRemoteRepo ? "cursor-pointer text-blue-500 hover:underline" : "")}
```
brendan-kellam, line 37: micro nit: usually when adding conditional CSS, we use `clsx`

packages/web/src/components/ui/dropdown-menu.tsx
```diff
@@ -47,7 +47,7 @@ const DropdownMenuSubContent = React.forwardRef<
   <DropdownMenuPrimitive.SubContent
     ref={ref}
     className={cn(
-      "z-50 min-w-[8rem] overflow-hidden rounded-md border bg-popover p-1 text-popover-foreground shadow-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+      "z-50 min-w-[6rem] overflow-hidden rounded-md border bg-popover p-1 text-popover-foreground shadow-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
```
brendan-kellam, line 50: Instead of modifying the component directly here, could we pass the `min-w-[6rem]` in the `className` property of the DropdownMenu?

packages/web/src/components/ui/data-table.tsx
```diff
@@ -37,100 +37,136 @@ export function DataTable<TData, TValue>({
...
+        getCoreRowModel: getCoreRowModel(),
+        getPaginationRowModel: getPaginationRowModel(),
+        getSortedRowModel: getSortedRowModel(),
+        getFilteredRowModel: getFilteredRowModel(),
+    });
 
     return (
-        <div>
-            <div className="flex items-center py-4">
+        <div className="pb-12">
```
brendan-kellam, line 65: nit: could we add a optional `className` props to `DataTableProps`, and then have `RepositoryTable` pass down this class?

msukkari:
closing this since repo table has changed significantly with v3, so not sure if this is applies anymore

PR #118, opened 2024-12-09
Add config option to specify max file size

brendan-kellam:
This PR adds the `settings.maxFileSize` option to the config schema, which allows users to specify the maximum file size (in bytes) that zoekt should index. E.g.,
```jsonc
{
    "settings": {
        "maxFileSize": 4194304 // Set size to 4MB (default 2MB)
    }
    // ...
}
```

All repositories in the config will use this file limit when indexing.

### DIFF
--- a/packages/backend/src/main.test.ts
+++ b/packages/backend/src/main.test.ts
@@ -79,7 +79,7 @@ test('isRepoReindexingRequired should return true when local excludedPaths chang
 
 test('isAllRepoReindexingRequired should return false when fileLimitSize has not changed', () => {
     const previous: Settings = {
-        fileLimitSize: 1000,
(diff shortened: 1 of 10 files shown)


packages/backend/src/main.test.ts
```diff
@@ -74,5 +74,9 @@ test('isRepoReindexingRequired should return true when local excludedPaths chang
         excludedPaths: ['node_modules', 'dist']
     };
 
-    expect(isRepoReindxingRequired(previous, current)).toBe(true);
+    expect(isRepoReindexingRequired(previous, current)).toBe(true);
 });
+
+test('isAllRepoReindexingRequired ...', () => {
```
brendan-kellam, line 80: UT

PR #128, opened 2024-12-13
Add `autoDeleteStaleRepos` config option

### LINKED ISSUE #39: GitLab repositories remain indexed after Sourcebot no longer has access to them
We have Sourcebot set up to index repositories hosted in a self-hosted GitLab instance. A dedicated Sourcebot account in GitLab is used to determine which repositories it has access to (by adding the user as a member of a repository or group) and only indexes those repositories.

However, if the Sourcebot user is removed as a member from a repository (or group) the repository remains in Sourcebot (presumably with a stale index).

What I would expect to happen is if Sourcebot no longer has access to a repository (or the repository has been deleted) then the repository should be removed from Sourcebot.

### LINKED ISSUE #127: GitLab topics are case insensitive
GitLab will display topics with capitals:

image

But will return them as normalized lowercase:
image

Since our current comparison util is case sensitive, this will fail:
```jsonc
{
    "repos": [
        {
            "type": "gitlab",
            "groups": [
                "sourcebot"
            ],
            "exclude": {
                "topics": [
                    "TypeScript" 
                ]
            }
        }
    ]
}
```

We should also be normalizing things to lowercase first (for GitLab).

brendan-kellam:
This PR adds the option `autoDeleteStaleRepos` to the `settings` object of the config. Its default is **true**. When enabled, stale repositories (i.e., the `.zoekt` index files in `.sourcebot/index`, and the checked out repository in `.sourcebot/repos`) are deleted from disk.

Fixes #39 

Also Fixes #127

### DIFF
--- a/packages/backend/src/db.ts
+++ b/packages/backend/src/db.ts
@@ -83,7 +83,7 @@ export const migration_addMaxFileSize = (schema: Schema, log?: (name: string) =>
 }
 
 /**
- * @todo: add PR link
+ * @see: https://github.com/sourcebot-dev/sourcebot/pull/128
  */
 export const migration_addDeleteStaleRepos = (schema: Schema, log?: (name: string) => void) => {
     if (schema.settings.autoDeleteStaleRepos === undefined) {
--- a/packages/backend/src/db.ts
+++ b/packages/backend/src/db.ts
@@ -13,13 +13,15 @@ export type Schema = {
     }
 }
 
+export const DEFAULT_DB_DATA: Schema = {
+    repos: {},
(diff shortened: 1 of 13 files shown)


packages/backend/src/db.ts
```diff
@@ -78,5 +79,17 @@ export const migration_addMaxFileSize = (schema: Schema, log?: (name: string) =>
         schema.settings.maxFileSize = DEFAULT_SETTINGS.maxFileSize;
     }
 
+    return schema;
+}
+
+/**
+ * @todo: add PR link
```
brendan-kellam, line 86: todo

PR #132, opened 2024-12-14
Refactor Language Support

KonradStanski:
This review centralizes the linguist language definitions.
I moves all language resolving methods to lib, separates the language syntax highlighting to a file which can be mapped to from the linguist metadata.
It also adds language syntax highlighting support for ~250 languages.


packages/web/package.json
```diff
@@ -46,19 +49,42 @@
...
+    "codemirror-lang-prolog": "^0.1.0",
+    "codemirror-lang-r": "^0.1.1",
+    "codemirror-lang-sparql": "^2.0.0",
+    "codemirror-lang-spreadsheet": "^1.3.0",
+    "codemirror-lang-zig": "^0.1.0",
     "embla-carousel-auto-scroll": "^8.3.0",
     "embla-carousel-react": "^8.3.0",
     "escape-string-regexp": "^5.0.0",
     "fuse.js": "^7.0.0",
+    "graphql": "^16.9.0",
```
brendan-kellam, line 87: do we need to take a dependency on graphql?

packages/web/src/lib/codemirrorLanguage.ts
```diff
@@ -0,0 +1,230 @@
...
+    "rust": rust(),
+    "sass": sass(),
+    "solidity": solidity,
+    "sparql": sparql(),
+    "spreadsheet": spreadsheet(),
+    "sql": sql(),
+    "svelte": svelte(),
+    "tsx": javascript({ jsx: true, typescript: true }),
+    "twig": twig(),
+    "typescript": javascript({ jsx: false, typescript: true }),
```
brendan-kellam, line 164: nit: do we need a `javascript` key that defines `javascript({ jsx: false, typescript: false })`?

packages/web/package.json
KonradStanski: It's a peer dependency for qraphql syntax highlighting.

packages/web/src/lib/codemirrorLanguage.ts
brendan-kellam: ah I see what you mean - I see that `JavaScript` is being mapped to `jsx` so no worries, you can disregard my comment.

Issue #14, opened 2024-10-01
FR: Support indexing code hosted on a local Bitbucket instance


gburiola:
https://confluence.atlassian.com/bitbucketserver

KonradStanski:
Does anyone know an example of a public bitcuket repo I could take a look at?
That way I can add the integration testing against some public instance.
Alternatively if someone has a private one they're willing to give me some auth credentials for testing purposes lmk.

KonradStanski:
https://bitbucket.org/atlassian/workspace/repositories/
Would this be an example? Or is this their cloud product and you are interested in self hosted only?
Perhaps for testing we need to set up:
https://confluence.atlassian.com/bitbucketserver/bitbucket-installation-guide-867338382.html

eddyg:
Bitbucket Data Center and Cloud are two _very_ different things. They have their similarities, but there are significant differences.

This request is specifically for _Data Center_ instances.

The Bitbucket Data Center API is documented here: [Bitbucket v9.3 API](https://developer.atlassian.com/server/bitbucket/rest/v903/intro/#about).

As you noted, you can spin-up your own developer instance of Bitbucket Data Center using [Atlassian's Docker image](https://hub.docker.com/r/atlassian/bitbucket).

brendan-kellam:
Thanks for inquiring @KonradStanski

@eddyg This is useful information thanks. It's pretty unfortunate that they have two offerings with almost the same name 😅

I need to read more into it, but my hope is that Data Center & Cloud have a shared api interface so that we don't need to care about which one the user is using.

eddyg:
For comparison/reference, here's the [Bitbucket Cloud API](https://developer.atlassian.com/cloud/bitbucket/rest/intro/#authentication)...

msukkari:
We've added support for Bitbucket Cloud and DC in v3.1.0! Instructions on how to configure here: https://docs.sourcebot.dev/docs/connections/bitbucket-data-center

PR #143, opened 2024-12-19
RFC: Bitbucket support

stevealx:
For issue #14 

I've tested this against bitbucket.org.

I was only able to test the server/data center support against a prism mock based on the OpenAPI spec.  I don't have
an instance to test against, so it may not work in real usage.  Hopefully, any fixes will be small.

The README file needs some screen shot updates for adding app passwords.  I may be able to sanitize one for Bitbucket Cloud, but a screenshot for the server version will need to come from someone who has it available.

This addresses my personal need, so I don't know how much more time I'm going to spend on it at present.  Maybe it will be useful to others, though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,4 @@ dist
 /bin
 /config.json
 .DS_Store
+.*.sw*
--- a/packages/backend/src/bitbucket.ts
+++ b/packages/backend/src/bitbucket.ts
@@ -1,0 +1,305 @@
+import { createBitbucketCloudClient } from "@coderabbitai/bitbucket/cloud";
+import { toBase64 } from "@coderabbitai/bitbucket";
+import { SchemaBranch, SchemaProject, SchemaRepository, SchemaTag, SchemaWorkspace } from "@coderabbitai/bitbucket/cloud/openapi";
+import { BitbucketConfig } from "./schemas/v2.js";
+import { excludeArchivedRepos, excludeForkedRepos, excludeReposByName, excludeReposByTopic, getTokenFromConfig, includeReposByTopic, marshalBool, measure } from "./utils.js";
+import { createLogger } from "./logger.js";
+import { AppContext, GitRepository } from "./types.js";
+import path from 'path';
+import micromatch from "micromatch";
+
+const logger = createLogger('Bitbucket');
+const BITBUCKET_CLOUD_HOSTNAME = 'bitbucket.org'; // note, not 'api.bitbucket.org'
+const BITBUCKET_CLOUD_API = 'https://api.bitbucket.org/2.0';
+
+interface Repo {
+    workspace: string;
+    project: string;
+    name: string;
+    isFork: boolean;
+    isArchived: boolean;
+    isPublic: boolean;
+    user: string;
+    token: string;
+    stars: number;
+    forks: number;
+};
+
+export const getBitbucketReposFromConfig = async (config: BitbucketConfig, ctx: AppContext) => {
+    const token = config.token ? getTokenFromConfig(config.token, ctx) : undefined;
+    const user = config.user ? config.user : undefined;
+    const basic = toBase64(user + ':' + token);
+    const url = config.url ? config.url : BITBUCKET_CLOUD_API;
+
+    logger.debug('BASE URL ' + url);
+
+    const clientOptions = {
+        baseUrl: url,
+        headers: {
+            Accept: 'application/json',
+            Authorization: `Basic ${basic}`,
+        },
+    };
+
+    const client = createBitbucketCloudClient(clientOptions);
+
+    const hostname = config.url ? new URL(config.url).hostname : BITBUCKET_CLOUD_HOSTNAME;
+
+    let foundRepos: Repo[] = [];
+
+    const workspaces = await getWorkspaces(client, url);
+    for (const workspace of workspaces) {
+        const projects = await getProjects(client, url, workspace);
+
+        for (const project of projects) {
+            const repos = await getRepos(client, url, workspace, project);
+
+            for (const repo of repos) {
+                logger.debug('REPO ' + JSON.stringify(repo));
+                if (repo.name == null) {
+                    continue;
+                }
+
+                var stars: number = 0;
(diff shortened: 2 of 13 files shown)


brendan-kellam:
Wow this is awesome!! Thanks for the contribution 😃

Will take a look at this later today. Will also take a look at getting a Bitbucket data center setup for testing. Also @gburiola, would you be willing to give this a try?

.gitignore
```diff
@@ -163,4 +163,5 @@ dist
 .sourcebot
 /bin
 /config.json
-.DS_Store
\ No newline at end of file
+.DS_Store
+.*.sw*
```
brendan-kellam, line 167: What are `.sw` files?

schemas/v2/index.json
```diff
@@ -516,6 +516,108 @@
...
+            "type": "object",
+            "properties": {
+                "type": {
+                    "const": "bitbucket",
+                    "description": "Bitbucket configuration"
+                },
+                "countMisc": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Count watchers and forks"
```
brendan-kellam, line 529: Still reading, but curious - what does this setting do?

stevealx: I added these counts for parity with the other services, although bitbucket doesn't seem to do stars the way github does.

Against bitbucket cloud I was seeing rate limiting of API calls when watchers and forks were counted for a large number of repos.  I made it configurable for the time being, since I don't need the info.

In the server API, I don't see these counts, so they will be zero in that case.

.gitignore
stevealx: I generally only use an IDE for Java.  These are state files that 'vim' creates when a file is being edited.  I don't like to see them in 'git status' output, and certainly would not want to check them in by accident.

Vim usage is why I added the .editorconfig file as well.  I didn't want tabs creeping into your code base.

packages/backend/src/bitbucket.ts
```diff
@@ -0,0 +1,643 @@
...
+                name: `${project.workspace}/${project.name}`,
+                id: repoId,
+                cloneUrl: cloneUrl.toString(),
+                path: repoPath,
+                isStale: false,
+                isFork: project.isFork,
+                isArchived: project.isArchived,
+                gitConfigMetadata: {
+                    //
+                    // using 'bitbucket-server' lets 'zoekt' generate the correct commit
```
brendan-kellam, line 155: To confirm, the URL pattern for bitbucket cloud is the same as bitbucket server?
Looking at [the zoekt code](https://github.com/sourcegraph/zoekt/blob/dc1b23bb0da9a0fa91056ad6d5fcf9b8e641f67d/gitindex/index.go#L146), it looks like the URL pattern is:
```javascript
// Commit URL:
// https://<bitbucketserver-host>/projects/<project>/repos/<repo>/commits/5be7ca73b898bf17a08e607918accfdeafe1e0bc

// File URL:
// https://<bitbucketserver-host>/projects/<project>/repos/<repo>/browse/<file>?at=5be7ca73b898bf17a08e607918accfdeafe1e0bc
```

brendan-kellam:
I didn't get a chance to go over everything since been out the last few days, so just going to release my feedback I had so far.

A general, code quality nit: we should consider making things stronger typed. For example, it would be great if we could type the response bodies s.t., we have more guarantees when we reach for certain properties.

I was hacking around a little bit, and _maybe_ something along these lines this could work

```TypeScript
import type { ClientOptions, ClientPathsWithMethod } from "openapi-fetch";
import { createBitbucketCloudClient } from "@coderabbitai/bitbucket/cloud";
import { SchemaProject as CloudProject } from "@coderabbitai/bitbucket/cloud/openapi";

// afaik, this is the only way of extracting the client API type
type CloudAPI = ReturnType<typeof createBitbucketCloudClient>;

// Defines a type that is a union of all API paths that have a GET method in the
// client api.
type CloudGetRequestPath = ClientPathsWithMethod<ClientAPI, "get">;

/**
* Defines a generic response body for any paginated api.
**/
type PaginatedResponse<T> = {
    readonly next?: string;
    readonly page?: number;
    readonly pagelen?: number;
    readonly previous?: string;
    readonly size?: number;
    readonly values?: readonly T[];
}


/**
* We need to do `V extends CloudGetRequestPath` since we will need to call `apiClient.GET(url, ...)`, which
* expects `url` to be of type `CloudGetRequestPath`. See example.
**/
const getPaginatedCloud = async <T, V extends CloudGetRequestPath>(baseUrl: V, get: (url: V) => Promise<PaginatedResponse<T>>) => {
    const results: T[] = [];
    let url = baseUrl;

    while (true) {
        const response = await get(url);

        if (!response.values || response.values.length === 0) { 
            break;
        }

        results.push(...response.values);

        if (!response.next) {
            break;
        }

        // cast required here since response.next is a string.
        url = response.next as V;
    }

    return results;
}
```

Usage example:
```TypeScript
async function cloudGetProjects(client: ClientAPI, workspace: string): Promise<string[]> {
    var results: string[] = [];

    const baseUrl = '/workspaces/{workspace}/projects';

    // `projects` is of type `CloudProject[]`.
    const projects = await getPaginatedCloud<CloudProject, typeof baseUrl>(baseUrl, async (url) => {
        
        // This response body will be typed to `/workspaces/{workspace}/projects`
        const response = await client.GET(url, {
            params: {
                path: {
                    workspace,
                }
            }
        });
        const { data, error } = response;
        if (error) {
            throw new Error (`Failed to fetch projects for workspace ${workspace}: ${error.type}`);
        }
        return data;
    });

    // --> All of this is now typed <--
    for (const project of projects) {
        if (project.key) {
            results.push(project.key);
        }
    }

    logger.debug(`  FOUND ${results.length} PROJECTS IN ${workspace}`);
    return results;
}
```

There are probably simpler ways of doing this so open to suggestions.

msukkari:
Closing since we've added support for this in v3: https://github.com/sourcebot-dev/sourcebot/pull/275

Appreciate the time you took to add this! If you get the chance to upgrade to v3 and try the new integration please let us know if you hit any issues

Issue #40, opened 2024-10-16
Sourcebot doesn't index the "default" branch if it has been changed


jerrykan:
We have noticed when using Sourcebot with our self-hosted GitLab instance that if the default branch of a repository changes Sourcebot doesn't start indexing the new default branch. Instead Sourcebot continues to index what was the default branch when the repository was first indexed.

What I would expect to happen if the default branch is changed (via `<repo> -> Settings -> Repository -> Branch defaults -> default Branch`) is that Sourcebot would index the new default branch the next time the repository is indexed.

I am aware of issue #29, but I believe this is a separate problem.

gaeljw:
For the record, I think I just hit the same issue or similar.

In my case I'm not entirely sure the default branch was changed in GitLab.

What I observed:
- we use both `master` and `my-default` branches, `my-default` is the default in GitLab, but both contain the same content at some point
- we stop using `master`
- months later, Sourcebot is still stuck on the `master` branch and do not index `my-default`

Surprinsingly, the file browser show content of `master` but when clicking on "Open in GitLab" do shows `my-default` branch (actually it shows `HEAD`).

Maybe for some reason Sourcebot was not able to find the default branch and used `master` as a default initially? (Or maybe we did switch the default branch but I doubt it).

EDIT: I cannot reproduce on a fresh installation of Sourcebot (the `my-default` branch do is indexed).

PR #148, opened 2024-12-26
Index default branch

Steven-Nagie:
Pertaining to issue #40, indexing default branch where available. 

Zoekt will index default branch forGitHub, GitLab, and Gitea projects, since those APIs include default branch info. 

Updated the GitRepository interface so the DB includes that info. Repositories will be marked for re-indexing if the default branch changes.


brendan-kellam:
Thanks for the PR! Sorry been behind on things - should be able to get to this today

Steven-Nagie:
@brendan-kellam let me know what you think.

gaeljw:
Would be nice to get this merged, anything blocker?

Steven-Nagie:
> Would be nice to get this merged, anything blocker?

Looks like there are some merge conflicts that would need to be worked through, but I could find time to work through them if I had some assurance that the PR would be reviewed.

brendan-kellam:
Heya - sorry for leaving this PR hanging for so long. Should have gotten this merged when it was first opened so that's my bad.

Re-reading this - getting the default branch from the code host makes sense as that will be hydrated with whatever it is updated to. I do wonder though if it makes sense to be indexing both HEAD and the default branch, or if we can just always index the default.

There are also some paths in search where we rely on `HEAD` in order to ensure zoekt will _not_ return results from multiple branches when there is no `rev` predicate in the search query ([here](https://github.com/sourcebot-dev/sourcebot/blob/4899c9fbc755851af2ddcce99f4a4200f2faa4f6/packages/web/src/features/search/searchApi.ts#L145)).

I have plans on refactoring some of the paths here anyways, so I think I will take this issue on and fold it with that. Appreciate the contribution and apologies again for taking so long

PR #158, opened 2025-01-14
v3 effort

brendan-kellam:
- Parallelized repo indexing
- Multi tenancy mode
- Authentication
- GUI connection management
- Nav bar indicators

https://docs.sourcebot.dev/docs/overview

PR #160, opened 2025-01-15
Add tenant ID concept into web app and backend

msukkari:
- tenantId into config schema, indexer will use this when calling zoekt-git-index
- store tenantId for repo in DB. This is needed atm because the indexer will pull repo object from DB to be able to index, and it needs the tenantId
- tenantId header (X-TENANT-ID) support in api/search that feeds it to zoekt

To run with MT enforced run `yarn dev:mt`

NOTE: For some reason passing in a tenantId of 0 to `/api/search` causes the tenantId to not be picked up in zoekt. As a result, if you've indexed a repo without a tenantId (defaults to 0) you'll be unable to search for them. To search against repos that don't have a tenant ID run without MT enforced (i.e. `yarn dev`)


brendan-kellam (approved):

schemas/v2/index.json
```diff
@@ -141,6 +141,9 @@
                         ["docs", "core"]
                     ]
                 },
+                "tenantId": {
```
brendan-kellam, line 144: Could you add a `@nocheckin` comment here? I'm using this tag to track things we should make sure to remove / change prior to shipping to main.
E.g., https://github.com/sourcebot-dev/sourcebot/pull/158/files#r1915644502

PR #163, opened 2025-01-15
parallelize repo indexing

msukkari:
This PR sets up a redis queue using BullMQ to parallelize repo indexing. High level changes:
- RepoIndexingStatus added to repo table in DB
- Main loop in backend will now fetch all the repos that should be indexed and adds them to the index queue. These repos are marked with the IN_INDEX_QUEUE status
- Workers pick up these index jobs and perform the git clone/fetch and indexing, and sets the status of the repo to INDEXED after its done
- Since fetching the git repos is an I/O heavy operation, we spin up NUM_CPU_CORES * indexConcurrencyMultiple workers. The default value of indexConcurrencyMultiple is set to 3 here. We can modify as needed from experimentation

The end result here is that new repos added to the DB are picked up almost immediately instead of being blocked on the backend to finish indexing the repos it may be indexing in that moment.


package.json
```diff
@@ -6,14 +6,20 @@
...
         "dev:backend": "yarn workspace @sourcebot/backend dev:watch",
-        "dev:web": "yarn workspace @sourcebot/web dev"
+        "dev:web": "yarn workspace @sourcebot/web dev",
+        "dev:redis": "docker ps --filter \"name=redis\" --format \"{{.Names}}\" | grep -q \"^redis$\" && docker rm -f redis; docker run -d --name redis -p 6379:6379 redis"
+
     },
     "devDependencies": {
         "npm-run-all": "^4.1.5"
+    },
+    "dependencies": {
```
brendan-kellam, line 21: nit: these dependencies should be placed in the `backend/package.json` since they are specific to the backend

brendan-kellam (approved):

PR #164, opened 2025-01-16
Authentication

brendan-kellam:
This PR adds authentication using the [auth.js](https://authjs.dev/) library. Changes are as follows:
- `auth.ts` configures the auth providers (e.g., GitHub), adapters (Prisma) and defines some other things.
- `middleware.ts` defines how requests should be handled. If a request is not authenticated, then we redirect to /login (or return a service error in the case of api routes).
- Added DB schemas for `User`,`Account`, and `VerificationToken`. These models are expected to exist by auth.js in order for the Prisma adapter to work. auth.js will take care of automatically creating entries in these tables when necessary. See [this document](https://authjs.dev/concepts/database-models)
- `login/page.tsx` is a basic login page

Next up:
- Add Me Control & migrate old settings dropdown there
- Start investigating authorization for multi-tenant

Some technical debt introduced in this PR:
- I've remove the ability to set the `DOMAIN_SUB_PATH` since it was causing issues with the matcher regex defined in `middleware.ts`. See comment in `entrypoint.sh` for more.


packages/web/src/lib/environment.ts
```diff
@@ -6,3 +6,7 @@ export const ZOEKT_WEBSERVER_URL = getEnv(process.env.ZOEKT_WEBSERVER_URL, "http
 export const SHARD_MAX_MATCH_COUNT = getEnvNumber(process.env.SHARD_MAX_MATCH_COUNT, 10000);
 export const TOTAL_MAX_MATCH_COUNT = getEnvNumber(process.env.TOTAL_MAX_MATCH_COUNT, 100000);
 export const NODE_ENV = process.env.NODE_ENV;
+
+export const AUTH_JS_SECRET = getEnv(process.env.AUTH_JS_SECRET); // Generate using `npx auth secret`
```
msukkari, line 10: nit: why not use the AUTH_SECRET convention? https://authjs.dev/reference/core/errors#missingsecret

packages/web/src/app/components/navigationMenu.tsx
```diff
@@ -1,90 +1,76 @@
...
-import { GitHubLogoIcon, DiscordLogoIcon } from "@radix-ui/react-icons";
-import { SettingsDropdown } from "./settingsDropdown";
 import { Separator } from "@/components/ui/separator";
 import Image from "next/image";
 import logoDark from "../../../public/sb_logo_dark_small.png";
 import logoLight from "../../../public/sb_logo_light_small.png";
-import { useRouter } from "next/navigation";
-
-const SOURCEBOT_DISCORD_URL = "https://discord.gg/6Fhp27x7Pb";
-const SOURCEBOT_GITHUB_URL = "https://github.com/sourcebot-dev/sourcebot";
```
msukkari, line 15: wouldn't this remove these links in the public search demo?

packages/web/src/app/components/navigationMenu.tsx
```diff
@@ -1,90 +1,76 @@
...
-                            size="icon"
-                            onClick={() => {
-                                window.open(SOURCEBOT_GITHUB_URL, "_blank");
-                            }}
+                            size="default"
                         >
-                            <GitHubLogoIcon className="w-4 h-4" />
+                            Logout
                         </Button>
-                        <SettingsDropdown />
```
msukkari, line 83: will the settings be readded in a future change?

packages/web/src/app/login/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+                                        // Docs:
+                                        // https://nextjs.org/docs/app/api-reference/functions/redirect#server-component
+                                        throw error
+                                    }
+                                }}
+                            >
+                                <Button
+                                    type="submit"
+                                >
+                                    {logo && (
```
msukkari, line 72: nit: github logo here is white against white button (im using dark mode, not sure if its better in light mode)

msukkari (approved):

packages/web/src/app/components/navigationMenu.tsx
brendan-kellam: Added these back

packages/web/src/app/components/navigationMenu.tsx
brendan-kellam: Added this back - in upcoming PR, I will probably add a MeControl dropdown that will contain these settings

PR #171, opened 2025-01-17
Multi tenancy support in config syncer

msukkari:


packages/backend/src/index.ts
```diff
@@ -19,7 +20,7 @@ type Arguments = {
 
 parser.add_argument("--configPath", {
     help: "Path to config file",
-    required: true,
+    required: false,
```
brendan-kellam, line 23: nit: could we make this param required if `SOURCEBOT_TENANT_MODE` is single?

brendan-kellam (approved):

PR #173, opened 2025-01-21
Organization switching & active org management

brendan-kellam:
This PR adds the following:
- A organization switcher which allows user to switch / create new organizations:
image

On the web server side, I'm trying out NextJS's [server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) to facilitate CRUD operations (in this case, org creation & org switching). These are more convenient than setting up new REST endpoints since type checking is basically automatic. Similar security concerns w.r.t., auth are the same though, so we will need to be careful when handling a request. See `actions.ts`.

- Added the concept of roles to the `UserToOrg` table. A user can either be a `OWNER`, granting them elevated access (e.g., ability to rename/delete the organization, billing, etc.), or a `MEMBER`.

- Introduced the concept of an "active organization" via the `activeOrgId` column in the `User` schema. The intention here is that a user will only ever be performing CRUD operations against a single organization at a time. We can use the `activeOrgId` on the user to determine what org they are interacting with for a given request. On User creation, a org is automatically created, the user is granted "OWNER" membership, and it's id is set to the activeOrgId.


packages/web/src/auth.ts
```diff
@@ -24,15 +32,69 @@ export const providerMap = providers
...
 export const { handlers, signIn, signOut, auth } = NextAuth({
     secret: AUTH_SECRET,
     adapter: PrismaAdapter(prisma),
     session: {
         strategy: "jwt",
     },
+    events: {
+        createUser: onCreateUser,
+    },
+    callbacks: {
```
brendan-kellam, line 85: Worth a comment here

packages/web/src/app/components/orgSelector/orgCreationDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    >
+                        <PlusCircledIcon className="h-5 w-5 text-muted-foreground" />
+                        Create organization
+                    </Button>
+                </DropdownMenuItem>
+            </DialogTrigger>
+            <DialogContent>
+                <DialogHeader>
+                    <DialogTitle>Create an organization</DialogTitle>
+                    <DialogDescription>Collaborate with</DialogDescription>
```
msukkari, line 57: nit: is this subtitle missing words?

packages/web/src/auth.ts
msukkari: could you explain what this doing? is this basically setting the user id from the token? if so, is this just syntax sugar vs. reading from the session token directly?

msukkari (approved):

packages/web/src/auth.ts
brendan-kellam: Yea basically what you said - it reads the JWT token's `sub` (subject) property (which happens to be the user's id) and writes it into `session.user`. The `session` object is sent over to the client.

Going to make things more explicit in a new revision - I don't think it's the best move to rely on the subject being set to the user id, so instead going to cache the user id in the JWT as a seperate field.

PR #174, opened 2025-01-22
Migrate to postgres

msukkari:
Migrates away from sqlite in favor of postgres. This is in preparation for secrets manager support, which typically requires a more robust DB solution.

- in dev mode, we spin up postgres docker container. This postgres instance will live on even after the app is spun down, and will mount to cache to persist DB across runs
- in the docker deployment we spin up postgres inside the container. This postgres instance will only live for the duration of the sourcebot docker container, but if the cache is mounted to the container the DB will persist

NOTE: due to DB folder ownership issues, a DB created in one deployment mode cannot be used by another. One must delete the DB before deploying using a different method else postgres will crash


Dockerfile
```diff
@@ -134,6 +139,7 @@ RUN chmod +x ./entrypoint.sh
 COPY default-config.json .
 
 EXPOSE 3000
+EXPOSE 5432
```
brendan-kellam, line 142: Do we need to expose this?

brendan-kellam (approved):

PR #177, opened 2025-01-24
Add `@sourcebot/schemas` package

brendan-kellam:
This PR centralizes our generated schema types & objects into the package `@sourcebot/schemas`. It exposes out the `yarn generate` command to read from the `schemas/` directory and generate a `.type.ts` and `.schema.ts` for each json file.


packages/schemas/src/v1/index.schema.ts
msukkari, line 1: should these generated files be checked in? is generation explicitly done on startup or is it expected that the user runs generate before starting for the first time?

msukkari (approved):

packages/schemas/src/v1/index.schema.ts
brendan-kellam: The files are generated  every time the package is installed (i.e., `yarn install` is run) because of the `postinstall` property in the `package.json` (which calls `yarn build`).

We don't technically need to check these files in, but it's mostly a convenience thing.

PR #180, opened 2025-01-25
add concept of secrets

msukkari:
This PR adds the concepts of secrets to Sourcebot. A secret is provided by the user and stored encrypted in the DB for later use. A config can reference a secret and the backend will automatically fetch it from the DB, decrypt it, and use it to clone the repo.

- In dev mode, you must set SOURCEBOT_ENCRYPTION_KEY, which is a 32 byte key. This can be stored in the `.env.local` file for both the backend and web app (added instructions to README)
- In the docker container, you can provide SOURCEBOT_ENCRYPTION_KEY as an env var. If one isn't provided, then we will create one and store it in `.sourcebot/.secret` for future use


packages/backend/src/main.ts
```diff
@@ -171,6 +206,13 @@ export const main = async (db: PrismaClient, context: AppContext) => {
                     { indexedAt: { lt: thresholdDate } },
                     { repoIndexingStatus: RepoIndexingStatus.NEW }
                 ]
+            },
```
brendan-kellam, line 209: unrelated to this PR, but thoughts on moving repo index management into a class?

packages/backend/src/main.ts
```diff
@@ -1,20 +1,47 @@
...
+        logger.error(`Repo ${repo.id} has no connections`);
+        return;
+    }
+
+    let token: string | undefined;
+    for (const repoConnection of repoConnections) {
+        const connection = repoConnection.connection;
+        const config = connection.config as unknown as ConnectionConfig;
+        if (config.token) {
+            token = await getTokenFromConfig(config.token, connection.orgId, db);
```
brendan-kellam, line 37: nit: should we early return here if we are wanting to get the first token?

packages/backend/src/main.ts
```diff
@@ -1,20 +1,47 @@
...
 import { ConnectionManager } from './connectionManager.js';
+import { ConnectionConfig } from '@sourcebot/schemas/v3/connection.type';
 
 const logger = createLogger('main');
 
-const syncGitRepository = async (repo: Repo, ctx: AppContext) => {
+type RepoWithConnections = Repo & { connections: (RepoToConnection & { connection: Connection})[] };
+
+// TODO: do this better? ex: try using the tokens from all the connections 
+// We can no longer use repo.cloneUrl directly since it doesn't contain the token for security reasons. As a result, we need to
```
brendan-kellam, line 21: Ah this is interesting, never considered this... using the first token makes sense. A repo belonging to multiple connections is likely a edge case anyways, so keeping it simple is a good tradeoff.

packages/backend/src/utils.ts
```diff
@@ -86,15 +87,39 @@ export const excludeReposByTopic = <T extends Repository>(repos: T[], excludedRe
     });
 }
 
-export const getTokenFromConfig = (token: string | { env: string }, ctx: AppContext) => {
+export const getTokenFromConfig = async (token: string | { env: string } | { secret: string }, orgId: number, db?: PrismaClient) => {
```
brendan-kellam, line 90: nit: you could probably use the `Token` type from `@sourcebot/schemas/v3/shared.type`

packages/db/prisma/schema.prisma
```diff
@@ -111,6 +112,19 @@ model UserToOrg {
   @@id([orgId, userId])
 }
 
+model Secret {
+  orgId               Int
+  key                 String
+  encryptedValue      String
+  iv                  String
```
brendan-kellam, line 119: what is iv?

packages/web/src/actions.ts
```diff
@@ -8,11 +8,142 @@ import { prisma } from "@/prisma";
...
+        where: {
+            orgId,
+        },
+        select: {
+            key: true,
+            createdAt: true
+        }
+    });
+
+    return secrets;
```
brendan-kellam, line 102: nit: to avoid accidentally leaking additional data, could we do `return secrets.map(secret => { key: secret.key, createdAt: secret.createdAt });`?

To mitigate if we ever make a whoopsies and delete the `select` clause in the findMany (say if we moved this to a `getSecrets(...)` function)

packages/web/src/actions.ts
```diff
@@ -8,11 +8,142 @@ import { prisma } from "@/prisma";
...
+        await prisma.secret.create({
+            data: {
+                orgId,
+                key,
+                encryptedValue: encrypted.encryptedData,
+                iv: encrypted.iv,
+            }
+        });
+    } catch (e) {
+        console.error(e);
```
brendan-kellam, line 56: Does printing / returning the exception have the potential of leaking something?

packages/web/src/app/secrets/secretsTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+        if ('keys' in keys) {
+            setSecrets(keys);
+        } else {
+            console.error(keys);
+        }
+    };
+
+    useEffect(() => {
+        fetchSecretKeys();
+    }, []);
```
brendan-kellam, line 36: nit: you could probably move the secret fetching into the parent component and pass in the data as props. This will make the secret fetching be done server side (since `SecretsPage` is a server component). This should solve the flicker that can be seen as the fetch occurs:

https://github.com/user-attachments/assets/bbca520f-3b93-4e9e-99c3-8511e6ae985a

packages/web/src/app/secrets/secretsTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+            });
+            return;
+        } else {
+            toast({
+                description: `✅ Secret created successfully!`
+            });
+        }
+
+        const keys = await getSecrets();
+        if ('keys' in keys) {
```
brendan-kellam, line 60: nit: you can use the `isServiceError` function as a type guard to determine if the response was a `ServiceError`

packages/web/src/app/secrets/secretsTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+
+    const keys = secrets.map((secret): SecretColumnInfo => {
+        return {
+            key: secret.key,
+            createdAt: secret.createdAt.toISOString(),
+        }
+    }).sort((a, b) => {
+        return new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime();
+    });
```
brendan-kellam, line 100: react nit: this would be a good place to use a [useMemo](https://react.dev/reference/react/useMemo) hook. It will make it s.t., `keys` is only re-computed when one if it's dependencies (`secrets`) changes, rather than on every re-render of the component.

brendan-kellam (approved):

packages/backend/src/main.ts
msukkari: yeah that would clean this up, added as backlog task

packages/web/src/actions.ts
msukkari: I don't think so but probably best to not return it just in case

packages/web/src/app/secrets/secretsTable.tsx
msukkari: fixed by passing in initial secrets set into child component, would that be an okay solution?

PR #181, opened 2025-01-28
enforce tenancy on search and repo listing endpoints

msukkari:
NOTE: this PR will break non-auth mode since it expects an auth session to exist for search and repoList endpoints!

- Run `yarn dev:mt` to run with tenancy enforced
- This will result in repos being indexed with tenancyId's, and all future searches and repo list calls to only return results for repos owned by the org

### DIFF
--- a/packages/web/src/app/api/(server)/search/route.ts
+++ b/packages/web/src/app/api/(server)/search/route.ts
@@ -15,10 +15,7 @@ export const POST = async (request: NextRequest) => {
 
     console.log(`Searching for org ${orgId}`);
     const body = await request.json();
-    const parsed = await searchRequestSchema.safeParseAsync({
-        ...body,
-        ...({orgId: orgId}),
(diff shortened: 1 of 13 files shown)


packages/web/src/app/api/(server)/search/route.ts
```diff
@@ -5,18 +5,19 @@ import { searchRequestSchema } from "@/lib/schemas";
...
+    }
 
+    console.log(`Searching for org ${orgId}`);
+    const body = await request.json();
     const parsed = await searchRequestSchema.safeParseAsync({
         ...body,
-        ...(tenantId ? {
-            tenantId: parseInt(tenantId)
-        } : {}),
+        ...({orgId: orgId}),
```
brendan-kellam, line 20: same nit about spreading in orgid

brendan-kellam (approved):

PR #184, opened 2025-02-10
add back gitlab, gitea, and gerrit support

msukkari:
NOTE: not super well tested (didnt test private repo case or self hosted case). Also missing local repo support (will need flag to figure out when to display this in new connections page)


packages/backend/src/connectionManager.ts
```diff
@@ -79,64 +78,44 @@ export class ConnectionManager implements IConnectionManager {
...
+                }
+                default: {
+                    return [];
                 }
-            })()
-        )
+            }
+        )();
+
+        // Filter out any duplicates by external_id and external_codeHostUrl.
```
brendan-kellam, line 101: Seems like this is duplicated below?

packages/backend/src/repoCompileUtils.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+export const compileGerritConfig = async (
+    config: GerritConnectionConfig,
+    connectionId: number,
+    orgId: number) => {
+
+    const gerritRepos = await getGerritReposFromConfig(config);
+    const hostUrl = config.url ?? 'https://gerritcodereview.com';
+    const hostname = config.url ? new URL(config.url).hostname : 'gerritcodereview.com';
```
brendan-kellam, line 163: nit: couldn't you just use `new URL(hostUrl).hostname`?

PR #191, opened 2025-02-14
enforce owner perms

msukkari:
Only owners can:
- send invite
- change billing email
- manage subscription

## Summary by CodeRabbit

- **New Features**
  - Introduced a new billing email update interface that validates inputs and provides real-time feedback.
  - Enhanced member management by enforcing owner-only permissions for inviting members and transferring ownership.
  - Updated subscription controls to clearly indicate when actions are restricted based on user role.

- **Refactor**
  - Strengthened role-based access across billing and membership settings for a more secure and intuitive user experience.


msukkari:
@coderabbitai review

packages/web/src/actions.ts
```diff
@@ -58,6 +58,37 @@ export const withOrgMembership = async <T>(session: Session, domain: string, fn:
     return fn(org.id);
 }
 
+export const withOwner = async <T>(session: Session, domain: string, fn: (orgId: number) => Promise<T>) => {
```
brendan-kellam, line 61: micro nit: since there is allot of overlap, we could maybe get away with single `withOrgMembership` with a optional param that specifies the minimum role. For example:

```
const withOrgMembership = async <T>(session: Session, domain: string, fn: (orgId: number) => Promise<T>, minimumRequiredRole?: OrgRole = OrgRole.MEMBER)
```

We did something similar in `requireOrgMembershipAndRole.ts` in monorepo

packages/web/src/actions.ts
```diff
@@ -574,6 +694,69 @@ export const fetchSubscription = (domain: string): Promise<Stripe.Subscription |
...
+            const userRole = await prisma.userToOrg.findUnique({
+                where: {
+                    orgId_userId: {
+                        orgId,
+                        userId: session.user.id,
+                    }
+                }
+            });
+
+            if (!userRole || userRole.role !== "OWNER") {
```
brendan-kellam, line 731: could you use `withOwner` here?

brendan-kellam (approved):

PR #194, opened 2025-02-14
cleanup org's repos and shards if it's inactive

msukkari:
This adds:
- A webhook to let stripe enform us when a subscription is created/deleted
- Garbage collection logic to cleanup an orgs repos/shards if it's been inactive (i.e. subscription is deleted) for 7 days


packages/backend/src/repoManager.ts
```diff
@@ -106,8 +106,33 @@ export class RepoManager implements IRepoManager {
...
+                stripeSubscriptionStatus: StripeSubscriptionStatus.INACTIVE,
+                stripeLastUpdatedAt: {
+                    lt: sevenDaysAgo
+                }
+            }
+        });
+
+        const inactiveOrgIds = inactiveOrgs.map(org => org.id);
+        
+        const inactiveOrgRepos = await this.db.repo.findMany({
```
brendan-kellam, line 121: Since inactive repos aren't necessarily orphaned, won't they be added back next time their corresponding connection syncs again? Do we need some logic to not sync a inactive org's connections?

brendan-kellam (approved):

packages/backend/src/repoManager.ts
msukkari: Added logic to mark org connections SYNC_NEEDED when they reactive

PR #197, opened 2025-02-17
Set resync and reindex period as environment variable

ErwanLeGallKosmos:
Hi 
I needed to be able to modify the ReSync and the ReIndex period value with environment variables.
You may also be interested in this.

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic configuration for background task intervals using environment settings, allowing for customizable timing with default values.
  
- **Chore**
	- Streamlined internal configuration management to improve maintainability and flexibility without altering user-facing functionality.  
	- Updated import statements to include new environment constants for better configuration handling.


msukkari:
Thanks for this! Out of curiosity was there a reason that you wanted to modify these values using an env var vs. the currently supported way of setting the values in the config? https://sourcebot.dev/search/browse/github.com/sourcebot-dev/sourcebot@HEAD/-/blob/packages/backend/src/main.ts?highlightRange=210%3A1%2C216%3A6

PR #201, opened 2025-02-19
add better visualization for connection/repo errors and warnings

msukkari:
- added indicators in nav bar for in progress jobs, connection warnings, and errors (failed repo indexing or failed connection sync)
- added more descriptive connection failed messages
- added warning/error indicators in connection list
- added failed repo index retry


packages/error/package.json
```diff
@@ -0,0 +1,13 @@
+{
```
brendan-kellam, line 1: I think you will need to update the Dockerfile as this is a new dependency for web / backend

packages/backend/src/github.ts
```diff
@@ -264,13 +348,43 @@ const getRepos = async (repoList: string[], octokit: Octokit, signal: AbortSigna
...
+    const failedResult = results.find(result => result.status === 'rejected');
+    if (failedResult && failedResult.status === 'rejected') {
+        throw failedResult.reason;
+    }
 
-            return [result.data];
-        } catch (e) {
-            logger.error(`Failed to fetch repository info for ${repo}.`, e);
-            throw e;
+    const validRepos: any[] = [];
```
brendan-kellam, line 374: nit: I see this pattern repeated, could we centralize into a shared function?

packages/backend/src/github.ts
```diff
@@ -264,13 +348,43 @@ const getRepos = async (repoList: string[], octokit: Octokit, signal: AbortSigna
             });
 
             logger.info(`Found info for repository ${repo} in ${durationMs}ms`);
+            return {
+                type: 'valid' as const,
+                data: result.data
+            };
+
+        } catch (error) {
+            if (error && typeof error === 'object' && 'status' in error && error.status === 404) {
```
brendan-kellam, line 357: nit: I see this pattern repeated, could we centralize into a shared function? Maybe `isNotFoundError`?

packages/backend/src/gitlab.ts
```diff
@@ -57,18 +67,41 @@ export const getGitLabReposFromConfig = async (config: GitlabConnectionConfig, o
...
+                        group
+                    };
+                }
                 throw e;
             }
-        }))).flat();
+        }));
 
-        allProjects = allProjects.concat(_projects);
+        const failedResult = results.find(result => result.status === 'rejected');
```
brendan-kellam, line 87: same nit about making this stuff shared

packages/backend/src/gitea.ts
```diff
@@ -197,18 +201,47 @@ const getReposOwnedByUsers = async <T>(users: string[], api: Api<T>) => {
...
+                    type: 'notFound' as const,
+                    user
+                };
+            }
             throw e;
         }
-    }))).flat();
+    }));
+
+    const failedResult = results.find(result => result.status === 'rejected');
```
brendan-kellam, line 220: same nit about sharing

packages/backend/src/github.ts
```diff
@@ -38,55 +41,78 @@ export const getGitHubReposFromConfig = async (config: GithubConnectionConfig, o
...
+        users: string[],
+        orgs: string[],
+        repos: string[],
+    } = {
+        users: [],
+        orgs: [],
+        repos: [],
+    };
+
+    try {
```
brendan-kellam, line 71: do we want to wrap everything in a try block here? I noticed that in other connection types we aren't doing this

packages/web/src/app/[domain]/connections/[id]/page.tsx
```diff
@@ -1,241 +1,236 @@
...
+import type { Connection, Repo, Org } from "@sourcebot/db"
+import { getConnectionInfoAction, getOrgFromDomainAction } from "@/actions"
+import { isServiceError } from "@/lib/utils"
+import { HoverCard, HoverCardContent, HoverCardTrigger } from "@/components/ui/hover-card"
+import { DisplayConnectionError } from "./components/connectionError"
+import { NotFoundWarning } from "./components/notFoundWarning"
+import { RetrySyncButton } from "./components/retrySyncButton"
+import { RetryAllFailedReposButton } from "./components/retryAllFailedReposButton"
 
 export default function ConnectionManagementPage() {
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/connections/[id]/page.tsx:32]: nit: looks like the tab size got messed up here. As an aside, we should really add linter rules at some point to enforce this stuff

packages/web/src/app/[domain]/components/warningNavIndicator.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const [warnings, setWarnings] = useState<Warning[]>([]);
+
+    useEffect(() => {
+        const fetchWarnings = async () => {
+            const connections = await getConnections(domain);
+            const warnings: Warning[] = [];
+            if (!isServiceError(connections)) {
+                for (const connection of connections) {
+                    const syncStatusMetadata = connection.syncStatusMetadata;
+                    if (syncStatusMetadata && typeof syncStatusMetadata === 'object' && ('notFound' in syncStatusMetadata)) {
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/components/warningNavIndicator.tsx:28]: nit: maybe for a follow up, but I'm seeing allot of these types of checks on `syncStatusMetadata` that enforce some structure of the underlying json blob.

Another approach could be to use [zod](https://zod.dev/) schema to define the structure for what we expect to be stored in the blob. This would act as the single source of truth for the structure of `syncStatusMetadata`. We could put this schema into a shared package (`@sourcebot/schemas` maybe?) s.t., both client & server have a shared view.

When you want to read the blob, you would do something like
```
const metadata = schema.safeParse(connection.syncStatusMetadata);
if (metadata.success) {
   const { notFound } = metadata;
   .... 
}
```

packages/web/src/actions.ts
```diff
@@ -785,6 +843,39 @@ export const getSubscriptionCheckoutRedirect = async (domain: string) =>
...
+                    id: connectionId,
+                },
+            });
+
+            if (!connection) {
+                return notFound();
+            }
+
+            const syncStatusMetadata = connection.syncStatusMetadata;
+            if (!syncStatusMetadata || typeof syncStatusMetadata !== 'object' || !('notFound' in syncStatusMetadata)) {
```
brendan-kellam, line 864: nit: see my related comment about zod

brendan-kellam (approved):

packages/backend/src/github.ts
msukkari: move to `connectionUtils.ts`

packages/backend/src/github.ts
msukkari: true, I forgot that I check this at a higher level - removed

PR #207, opened 2025-02-22
Inline secret creation

brendan-kellam:
## Secret combobox:

Empty:
image

Token set:
image

Dropdown:
image

## Import a secret dialog:
image


packages/schemas/src/v3/connection.schema.ts
```diff
@@ -62,6 +62,7 @@ const schema = {
             "type": "string",
             "pattern": "^[\\w.-]+$"
           },
+          "default": [],
```
msukkari, line 65: nit: is there a reason we're doing this for only orgs and not the other fields?

packages/schemas/src/v3/github.schema.ts
```diff
@@ -58,6 +58,7 @@ const schema = {
         "type": "string",
         "pattern": "^[\\w.-]+$"
       },
+      "default": [],
```
msukkari, line 61: same as previous comment

packages/schemas/src/v3/connection.schema.ts
msukkari: nevermind I see that this is the generated file and the schema adds them to all of them, so probably just an old version of the generated files or something

msukkari (approved):

PR #208, opened 2025-02-24
add posthog events on various user actions

msukkari:


packages/web/src/app/[domain]/components/configEditor.tsx
```diff
@@ -160,7 +171,7 @@ const ConfigEditor = <T,>(props: ConfigEditorProps<T>, forwardedRef: Ref<ReactCo
                                         onClick={(e) => {
                                             e.preventDefault();
                                             if (editorRef.current?.view) {
-                                                onQuickAction(fn, value, editorRef.current.view, {
+                                                onQuickAction(fn, captureEvent, name, type, value, editorRef.current.view, {
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/components/configEditor.tsx:174]: nit: could we just have the `wa_config_editor_quick_action_pressed` event here instead of having to plumb the captureEvent function to the onQuickAction callback?

packages/web/src/app/[domain]/onboard/page.tsx
```diff
@@ -54,12 +54,9 @@ export default async function Onboard({ params, searchParams }: OnboardProps) {
                     <ConnectCodeHost
                         nextStep={OnboardingSteps.InviteTeam}
                     />
-                    <Link
-                        className="text-sm text-muted-foreground underline cursor-pointer mt-12"
-                        href={`?step=${lastRequiredStep}`}
-                    >
-                        Skip onboarding
-                    </Link>
+                    <SkipOnboardingButton
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/onboard/page.tsx:57]: nit: could you pass `lastRequiredStep` as a prop? Since when we get to the self-hosting case, the last step won't always be checkout

packages/web/src/app/[domain]/connections/components/connectionList/connectionListItemManageButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const captureEvent = useCaptureEvent()
+
+    return (
+        <Button
+            variant="outline"
+            size={"sm"}
+            className="ml-4"
+            onClick={() => {
+                captureEvent('wa_connection_list_item_manage_pressed', {})
+                window.location.href = `connections/${id}`
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/connections/components/connectionList/connectionListItemManageButton.tsx:22]: nit: I'm not sure it really matters, but the convention I've been using so far is to use the NextJs router for redirecting (i.e., `router.push`). See: https://nextjs.org/docs/pages/building-your-application/routing/redirecting

brendan-kellam (approved):

PR #215, opened 2025-02-27
Add back revision support

brendan-kellam:


packages/backend/src/zoekt.ts
```diff
@@ -1,17 +1,58 @@
...
+
+        revisions = [
+            ...revisions,
+            ...matchingTags
+        ];
+    }
+
+    // zoekt has a limit of 64 branches/tags to index.
+    if (revisions.length > 64) {
+        logger.warn(`Too many revisions (${revisions.length}) for repo ${repo.id}, truncating to 64`);
```
msukkari, line 52: can you add a posthog event for this case?

msukkari (approved):

PR #220, opened 2025-02-28
revamp repo page

msukkari:
image

image
image


packages/web/src/actions.ts
```diff
@@ -325,12 +325,25 @@ export const getRepos = async (domain: string, filter: { status?: RepoIndexingSt
                 }
             });
 
+            const connectionData = await prisma.connection.findMany({
```
brendan-kellam, line 328: nit: you could probably change the `prisma.repo.findMany` query to include the actual connection (rather than just the join table), that way you don't need to do this second query.

Something like:
```typescript
const repos = await prisma.repo.findMany({
      where: {
          orgId,
          ...(filter.status ? {
              repoIndexingStatus: { in: filter.status }
          } : {}),
          ...(filter.connectionId ? {
              connections: {
                  some: {
                      connectionId: filter.connectionId
                  }
              }
          } : {}),
      },
      include: {
          connections: {
              include: {
                  connection: true,
              }
          }
      }
  });
```

brendan-kellam:
Could add a `Repositories` header, similar to what we are doing with Connections and Settings? That way these pages are more cohesive.

image
image

packages/web/src/app/[domain]/repos/page.tsx
```diff
@@ -13,11 +12,9 @@ export default async function ReposPage({ params: { domain } }: { params: { doma
     return (
         <div className="h-screen flex flex-col items-center">
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/repos/page.tsx:13]: Similar comment to the header one: to make things cohesive, we should polly use `bg-backgroundSecondary` here

brendan-kellam (approved):

PR #222, opened 2025-03-01
switch magic link to invite code

msukkari:
Previously our magic link behavior would be to provide a link that would automatically sign the user in. Unfortunately, some email scanners will visit this link and invalidate the auth token. As a result, let's switch to using an invite code that the user enters to complete sigin in.

NOTE: we are storing the encrypted user email here as a cookie because we need to read it in the `login/verify` page, as it's a required param to authjs. I tried to figure out how to do this as a param (i.e. `login/verify?email=<email`) but I couldn't get it to work. It seems like this redirect is hardcoded with authjs (it will take it to `verifyRequest` internally) and isn't configurable.


packages/web/src/app/[domain]/components/settingsDropdown.tsx
```diff
@@ -64,7 +64,11 @@ export const SettingsDropdown = ({
     }, [theme]);
 
     return (
-        <DropdownMenu>
+        <DropdownMenu onOpenChange={(isOpen) => {
```
Comment (msukkari) [inline packages/web/src/app/[domain]/components/settingsDropdown.tsx:67]: for some reason I was hitting a bug where the first time we sign in using magic link the settings drop down wouldn't have a valid session. Not sure if this is a new bug or if it existed in the previous magic link logic. Basic fix is to just update the session every time they press the settings dropdown (this isn't a common interaction and this is light weight, so should be fine)

Comment (brendan-kellam) [reply packages/web/src/app/[domain]/components/settingsDropdown.tsx:67]: Sounds good - could you add this a comment?

PR #233, opened 2025-03-20
Single tenancy & auth modes

brendan-kellam:
This change introduces two new environment variables:

- `SOURCEBOT_AUTH_ENABLED` controls whether authentication is enabled. It can be set to `true` or `false`.
- `SOURCEBOT_TENANCY_MODE` controls the tenancy mode of Sourcebot. It can be set to `single` or `multi`.

The following breaks down the configuration matrix for these options:
image

In a upcoming PR, we will add support for declarative connection configs (via a `config.json`) when in single tenant mode.


packages/web/src/env.mjs
```diff
@@ -36,6 +37,9 @@ export const env = createEnv({
         NODE_ENV: z.enum(["development", "test", "production"]),
         SOURCEBOT_TELEMETRY_DISABLED: booleanSchema.default('false'),
         DATABASE_URL: z.string().url(),
+
+        SOURCEBOT_TENANCY_MODE: tenancyModeSchema.default("multi"),
```
brendan-kellam, line 41: Consider: maybe we should make the tenancy mode default to single since it's most-likely the golden path for self-hosted users anyways.

CC @msukkari

PR #259, opened 2025-04-03
Display name improvements

brendan-kellam:
This PR does the following:
- Adds an optional `displayName` field to the Repos table. This field differs to `name` in that it does not include the `hostUrl` as a prefix, and thus is shorter & easier to read. The displayName is also passed along to zoekt via `zoekt.display-name`. This field replaces needing to create a display name client side.
- Altered the behaviour in `RepoManager::syncGitRepository` to always "upsert" the repo's `metadata.gitConfig` regardless of clone or fetch (previously it only happened on fetch). This makes it s.t., the data in the database (`metadata.gitConfig`) is **eventually consistent** with the git config on the repo. This is essential for when we add new fields (like `zoekt.display-name` in this case).

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved formatting issues for repository names (notably for Gerrit), resulting in clearer and more consistent display in the interface.

- **New Features**
  - Standardized repository naming across the platform, enabling full names to be displayed reliably in listings and carousels with a dedicated display label for improved user clarity.

### DIFF
--- a/packages/backend/src/git.ts
+++ b/packages/backend/src/git.ts
@@ -1,21 +1,15 @@
 import { simpleGit, SimpleGitProgressEvent } from 'simple-git';
 
-export const cloneRepository = async (cloneURL: string, path: string, gitConfig?: Record<string, string>, onProgress?: (event: SimpleGitProgressEvent) => void) => {
+export const cloneRepository = async (cloneURL: string, path: string, onProgress?: (event: SimpleGitProgressEvent) => void) => {
     const git = simpleGit({
         progress: onProgress,
     });
-
-    const configParams = Object.entries(gitConfig ?? {}).flatMap(
-        ([key, value]) => ['--config', `${key}=${value}`]
-    );
-
     try {
         await git.clone(
             cloneURL,
             path,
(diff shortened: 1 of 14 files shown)


packages/backend/src/git.ts
```diff
@@ -48,6 +42,26 @@ export const fetchRepository = async (path: string, onProgress?: (event: SimpleG
...
+ * present in gitConfig.
+ */
+export const upsertGitConfig = async (path: string, gitConfig: Record<string, string>, onProgress?: (event: SimpleGitProgressEvent) => void) => {
+    const git = simpleGit({
+        progress: onProgress,
+    }).cwd(path);
+
+    try {
+        for (const [key, value] of Object.entries(gitConfig)) {
+            await git.addConfig(key, value);
```
msukkari, line 58: random question but how does state work here? is this applying this config to some global simple-git state?

brendan-kellam: Discussed offline: this applies to the repo's local git config state. https://git-scm.com/docs/git-config

PR #267, opened 2025-04-10
Only create DB directory if using embedded DB

jerrykan:
If an external database is being used there is no need to create the local database files. This prevents the database files from being created unless the embedded database is being used.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the database initialization and startup process by centralizing configuration checks, ensuring more consistent handling of embedded database settings.


msukkari (approved):

msukkari:
Thanks! Will merge tmwr when I get the chance to sanity check repro

Issue #270, opened 2025-04-23
Failed to clone repository


mstimvol:
When Sourcebot syncs my repositories (around 480), a few of them (8) are failing with the following error:

> [backend] | 2025-04-23T15:51:19.622Z error: [RepoManager] Failed to sync repository 447, attempt 2/3. Sleeping for 10s... Error: Error: Failed to clone repository

My conf looks like:

```
{
    "$schema": "https://raw.githubusercontent.com/sourcebot-dev/sourcebot/refs/heads/main/schemas/v3/index.json",
    "connections": {
        "some-connection1": {
            "type": "github",
            "repos": ["myorga/repo"],
            "revisions": {
                "branches": [
                    "main",
                    "stable_v1",
                    "stable_v2"
                ]
            }
        },
        "some-connection2": {
            "type": "github",
            "repos": ["Behat/Gherkin"]
        },
        "some-connection3": {
            "type": "gitlab",
            "token": {
                "env": "GITLAB_TOKEN"
            },
            "groups": [
                "plugins"
            ],
            "url": "https://my-self-hosted-instance.tld"
        }
    }
}
```

I first thought it's because of the API limit in Gitlab or because one or two projects are archived, but I figured out that it also fails to sync sometimes "some-connection1" and "some-connection2" and furthermore, the 8 repositories are always other repos - not the same. Any ideas why the indexing fails?

I'm using v3.0.4 of Sourcebot

msukkari:
Hey Tim, sorry you're hitting this. Since the specific repos that are failing is changing I suspect this might be a networking issue when cloning the repos (I'm assuming the fail is happening during cloning not indexing from your error log). We've hit this before in certain environments where the networking is limited.

A workaround I would suggest you try is to drop the `maxRepoIndexingJobConcurrency` in the config: https://demo.sourcebot.dev/~/browse/github.com/sourcebot-dev/sourcebot@HEAD/-/blob/schemas/v3/index.json?highlightRange=46%3A1%2C50%3A19

Let me know if that helps!

mstimvol:
Hi Michael, changing `maxRepoIndexingJobConcurrency` from 8 to 2 finally fixed the error. Thank you very much!

securisec:
@msukkari having the same issue with gitea. Here is my quick test setup. The user `admin` has 4 repos in gitea. 

docker-compose.yml
```yaml
version: "3"

networks:
  gitea:
    external: false

services:
  gitea:
[...]

PR #303, opened 2025-05-10
Align gitea clone_url with gitea host url. Addresses #270

securisec:
This PR aligns gitea cloneUrl with hostUrl. When gitea is deployed as part of a docker-compose stack, it will incorrectly get the `cloneUrl` from the gitea API which will indicate `localhost` which is not correct. This will case sourcebot to fail when cloning the repo. 

This fix ensures that the user specified `host` url is always used when cloning repos also. Addresses #270 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved consistency of repository clone URLs by ensuring their host matches the configured host URL for Gitea repositories.


msukkari (approved):

msukkari:
Thanks!

PR #304, opened 2025-05-11
[Docs] SOURCEBOT_TELEMETRY_DISABLED should be set as true

ittaiz:
https://docs.sourcebot.dev/self-hosting/overview

## Summary by CodeRabbit

- **Documentation**
  - Clarified the correct environment variable value for disabling telemetry in the documentation.


brendan-kellam (approved):
thx for the catch!

PR #335, opened 2025-06-06
Adds support for encrypted license keys

msukkari:
## Summary by CodeRabbit

- **New Features**
  - Added cryptographic signature verification for license keys to enhance security.
  - Introduced support for specifying a public key path via environment variables.

- **Improvements**
  - License expiry date now displays both date and time, including time zone information, for clearer license status visibility.

- **Chores**
  - Updated environment and build configurations to support new public key requirements.
  - Added documentation for the new public key environment variable.

### DIFF
--- a/packages/crypto/src/index.ts
+++ b/packages/crypto/src/index.ts
@@ -5,6 +5,8 @@ import { SOURCEBOT_ENCRYPTION_KEY } from './environment';
 const algorithm = 'aes-256-cbc';
 const ivLength = 16; // 16 bytes for CBC
 
+const publicKeyCache = new Map<string, string>();
+
 const generateIV = (): Buffer => {
     return crypto.randomBytes(ivLength);
 };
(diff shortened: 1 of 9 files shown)


packages/web/src/env.mjs
```diff
@@ -55,6 +55,8 @@ export const env = createEnv({
 
         DATA_CACHE_DIR: z.string(),
 
+        SOURCEBOT_PUBLIC_KEY_PATH: z.string().optional(),
```
brendan-kellam: Could we make this environment variable required? The public key should always be available I assume?

packages/crypto/src/index.ts
```diff
@@ -63,3 +66,27 @@ export function decrypt(iv: string, encryptedText: string): string {
 
     return decrypted;
 }
+
+export function verifySignature(data: string, signature: string, publicKeyPath: string): boolean {
+    try {
+        let publicKey = publicKeyCache.get(publicKeyPath);
```
brendan-kellam, line 72: micro nit: couldn't we just load the public key once in `entitlements/server.ts` and then pass the `publicKey` to this function?

packages/crypto/src/index.ts
```diff
@@ -63,3 +66,27 @@ export function decrypt(iv: string, encryptedText: string): string {
...
+        if (!publicKey) {
+            if (!fs.existsSync(publicKeyPath)) {
+                throw new Error(`Public key file not found at: ${publicKeyPath}`);
+            }
+            
+            publicKey = fs.readFileSync(publicKeyPath, 'utf8');
+            publicKeyCache.set(publicKeyPath, publicKey);
+        }
+
+        const base64Signature = signature.replace(/-/g, '+').replace(/_/g, '/');
```
brendan-kellam, line 84: nit: could you add a comment explaining what is going on here?

brendan-kellam:
Could you add a changelog entry?

brendan-kellam (approved):

packages/crypto/src/index.ts
msukkari: trying to keep data related to encrypt/decrypt in the crypto package

PR #340, opened 2025-06-06
[search/browse] link repo name to file browser; link code image to external

drew-u410:
* [Change] Updates repository name to use the file browser / tree by default instead of the external repository.
* [New] Adds a link to the external repository on the source code logo.

## Summary by CodeRabbit

- **New Features**
  - Repository icon now links directly to the repository if a link is available.
- **Refactor**
  - Repository name is always styled as clickable and navigates to the repository root path on click, replacing the previous external link behavior.

### DIFF
--- a/packages/web/src/app/[domain]/components/pathHeader.tsx
+++ b/packages/web/src/app/[domain]/components/pathHeader.tsx
@@ -223,22 +223,27 @@ export const PathHeader = ({
     return (
         <div className="flex flex-row gap-2 items-center w-full overflow-hidden">
             {info?.icon ? (
-                <Image
-                    src={info.icon}
-                    alt={info.codeHostName}
-                    className={`w-4 h-4 ${info.iconClassName}`}
-                />
-            ): (
+                <a href={info.repoLink} target="_blank" rel="noopener noreferrer">
+                    <Image
+                        src={info.icon}
+                        alt={info.codeHostName}
+                        className={`w-4 h-4 ${info.iconClassName}`}
+                    />
+                </a>
+            ) : (
                 <LaptopIcon className="w-4 h-4" />
             )}
-            <Link
-                className={clsx("font-medium", {
-                    "cursor-pointer hover:underline": info?.repoLink,
+            <div
+                className="font-medium cursor-pointer hover:underline"
+                onClick={() => navigateToPath({
+                    repoName: repo.name,
+                    path: '',
+                    pathType: 'tree',
+                    revisionName: branchDisplayName,
                 })}
-                href={info?.repoLink ?? ""}
             >
                 {info?.displayName}
-            </Link>
+            </div>
             {branchDisplayName && (
                 <p
                     className="text-xs font-semibold text-gray-500 dark:text-gray-400 mt-[3px] flex items-center gap-0.5"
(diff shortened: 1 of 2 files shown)


packages/web/src/app/[domain]/components/pathHeader.tsx
```diff
@@ -223,19 +225,29 @@ export const PathHeader = ({
...
+                        className={`w-4 h-4 ${info.iconClassName}`}
+                    />
+                )
+            ) : (
                 <LaptopIcon className="w-4 h-4" />
             )}
             <Link
                 className={clsx("font-medium", {
-                    "cursor-pointer hover:underline": info?.repoLink,
+                    "cursor-pointer hover:underline": true,
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/components/pathHeader.tsx]: nit: we can remove this `clsx` since this is no longer conditional

packages/web/src/app/[domain]/components/pathHeader.tsx
```diff
@@ -223,19 +225,29 @@ export const PathHeader = ({
...
                 })}
-                href={info?.repoLink ?? ""}
+                href={`/${domain}/browse/${repo.name}@HEAD/-/blob/`}
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/components/pathHeader.tsx]: We have a hook for browse navigations if you take a look at `useBrowseNavigation.ts` . May need to make this a a div with a `onClick` callback

packages/web/src/lib/utils.ts
```diff
@@ -174,6 +175,7 @@ export const getCodeHostInfoForRepo = (repo: {
                 repoLink: webUrl,
                 icon: src,
                 iconClassName: className,
+                webUrl,
```
brendan-kellam: Don't think we actually need to add this field since `repoLink` exists?

brendan-kellam:
Also if you could add a changelog entry 🙏

packages/web/src/lib/utils.ts
drew-u410: Yup!  Sorry- was iterating w/ternary on a env var and didn't clean this up!

drew-u410:
> Also if you could add a changelog entry 🙏

Added!

I think all good now if you want to give it a local test.

packages/web/src/app/[domain]/components/pathHeader.tsx
```diff
@@ -223,22 +223,27 @@ export const PathHeader = ({
...
             )}
-            <Link
-                className={clsx("font-medium", {
-                    "cursor-pointer hover:underline": info?.repoLink,
+            <div
+                className="font-medium cursor-pointer hover:underline"
+                onClick={() => navigateToPath({
+                    repoName: repo.name,
+                    path: '',
+                    pathType: 'blob',
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/components/pathHeader.tsx]: nit: I think this `pathType` will need to be `tree`

Comment (brendan-kellam) [reply packages/web/src/app/[domain]/components/pathHeader.tsx]: Yea seems like it. Navigating to the root with `blob` will pick a random file at the root (this is a symptom of how we use [zoekt to get the file source](https://github.com/sourcebot-dev/sourcebot/blob/main/packages/web/src/features/search/fileSourceApi.ts), which we are planning on refactoring soon for unrelated reasons.)

image

Comment (drew-u410) [reply packages/web/src/app/[domain]/components/pathHeader.tsx]: Oh neat!  Updated.

brendan-kellam (approved):
LGTM 👍

PR #354, opened 2025-06-16
fix(search-contexts): Fix issue where a repository would not appear in a search context if it was created after the search context was created

brendan-kellam:
## Problem

If a repository is added **after** a search context (e.g., a new repository is synced from the code host), then it will never be added to the context even if it should be included. The workaround is to restart the instance.

## Solution

This PR adds a call to re-sync all search contexts whenever a connection is successfully synced. This PR adds the `@sourcebot/shared` package that contains `syncSearchContexts.ts` (previously in web) and it's dependencies (namely the entitlements system).

## Why another package?

Because the `syncSearchContexts` call is now called from:
1. `initialize.ts` in **web** - handles syncing search contexts on startup and whenever the config is modified in watch mode. This is the same as before.
2. `connectionManager.ts` in **backend** - syncs the search contexts whenever a connection is successfully synced.

## Follow-up devex work
Two things:
1. We have several very thin shared packages (i.e., `crypto`, `error`, and `logger`) that we can probably fold into this "general" shared package. `schemas` and `db` _feels_ like they should remain separate (mostly because they are "code-gen" packages).
2. When running `yarn dev`, any changes made to the shared package will only get picked if you `ctrl+c` and restart the instance. Would be nice if we have watch mode work across package dependencies in the monorepo.

## Summary by CodeRabbit

- **New Features**
  - Introduced a shared package (`@sourcebot/shared`) providing centralized constants, environment configuration, entitlement utilities, and configuration loading for both backend and frontend.
  - Added new utility functions for Base64 decoding, remote path detection, and configuration validation.

- **Refactor**
  - Migrated entitlements logic, plan definitions, and related utilities from local modules in web and backend packages to the shared package.
  - Updated imports throughout the codebase to use the new shared package for entitlements, constants, and configuration utilities.
  - Refactored configuration loading and validation in both backend and web to use the shared utility.
  - Enhanced synchronization of search contexts with organization-specific context and database parameters.

- **Chores**
  - Updated Dockerfile and Makefile to include and clean shared package dependencies and build artifacts.
  - Added and updated package metadata and TypeScript configuration for the shared package.
  - Updated license to cover new shared directories.

- **Bug Fixes**
  - Improved reliability and consistency of entitlement checks and configuration loading by centralizing logic.
  - Fixed issue where repositories created after a search context were not included in that context.

- **Style**
  - Minor changes such as adding trailing newlines and updating logger labels for clarity.

- **Revert**
  - Removed redundant or now-centralized local entitlements and constants files.


packages/shared/src/env.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const env = createEnv({
+    server: {
+        SOURCEBOT_EE_LICENSE_KEY: z.string().optional(),
+        SOURCEBOT_PUBLIC_KEY_PATH: z.string(),
+        NEXT_PUBLIC_SOURCEBOT_CLOUD_ENVIRONMENT: z.enum(SOURCEBOT_CLOUD_ENVIRONMENT).optional(),
+    },
+    runtimeEnv: process.env,
+    emptyStringAsUndefined: true,
+    // @note: todo
```
brendan-kellam: todo

packages/shared/src/index.server.ts
msukkari, line 1: out of curiosity why are we adding separate index files for server/client here?

msukkari (approved):

packages/shared/src/index.server.ts
brendan-kellam: Good question (and something I'll add to a README).

`index.server.ts` includes Node.JS specific imports (like the `fs` package). When you try to import it on the client, you hit something like this:
image

There are some things we would like to use on the client (namely constants), so having separate index files was my _ugly_ solution.

PR #365, opened 2025-06-25
fix(browse): Fix issue where files would sometimes never load

brendan-kellam:
## Problem

We noticed that sometimes files would be stuck in a infinite loading state like the following:

image

This would happen randomly. Usually refreshing the page would fix it.

## Solution

This PR moves the file source fetch from the client to the server. The downside of this approach is that all of the **client side** prefetching we were doing (with `usePrefetchFolderContents.ts` and `usePrefetchFileSource.ts`) no longer works, so now users will always see the loading spinner.

We should probably switch to links anyways (see #353), so we can probably rely on Next's [built-in prefetching with the Link component.](https://nextjs.org/docs/app/api-reference/components/link#prefetch).

## Why was this happening in the first place?

TLDR: I'm not _totally sure_, but I think it's because we are using server actions incorrectly. See [this issue](https://github.com/TanStack/query/issues/7934).

I did some digging, and came across [this issue](https://github.com/TanStack/query/issues/7934) that mentioned that server actions are not intended for fetching data and that you can get some unexpected behaviours when you do. This looks to be the case according to the [react docs](https://react.dev/reference/rsc/use-server#caveats). It seems like this is one of those cases, but I'm not exactly sure why. I confirmed that this was the issue by switching over to using a `fetch` call and was unable to repro. Moving the data fetching to the server also fixed the issue.

As a longer-term thing, it might make sense to move away from using server actions (either altogether or for specifically data fetching) to avoid issues like this. Ideally something with as good of a DX (or better). Some options I've come across:
- https://trpc.io/
- https://ts-rest.com/

## Summary by CodeRabbit

- **New Features**
  - Introduced a new PureTreePreviewPanel component for displaying file tree previews.

- **Refactor**
  - Converted several components (BrowsePage, CodePreviewPanel, TreePreviewPanel) from client-side to async server components, streamlining data fetching and removing client-side state management.
  - Simplified the useBrowseParams hook by delegating path parsing logic to a new utility function.

- **Bug Fixes**
  - Improved error handling and eliminated redundant loading states in preview panels.

- **Chores**
  - Removed prefetch-on-hover functionality and related hooks across file tree, breadcrumbs, and reference list components, reducing unnecessary background data fetching.
  - Deleted unused hooks for prefetching file sources and folder contents.


msukkari (dismissed):

brendan-kellam:
[edit]: resolved in https://github.com/sourcebot-dev/sourcebot/pull/365/commits/417eabf718090590d3efb87b80c7defbf0ffa7ad

todo: fix bug with navigating to repository root.
image

Issue #263, opened 2025-04-04
[bug] Unable to upgrade to v3


gaeljw:
### Describe the bug

Hello folks,

I just tried to upgrade to v3 (from 2.8.4) following https://docs.sourcebot.dev/self-hosting/upgrade/v2-to-v3-guide but upon startup of the v3, I've got errors related to Postgres.

I haven't had time to investigate yet, this may be a permission issue on our end.

In doubt, I thought I'd share it :)

Logs during startup:
```
[Info] Waiting for the database to be ready...
postgres: could not access the server configuration file "/data/.sourcebot/db/postgresql.conf": No such file or directory
localhost:5432 - no response
[Info] Waiting for the database to be ready...
localhost:5432 - no response
[Info] Waiting for the database to be ready...
localhost:5432 - no response
[Info] Waiting for the database to be ready...
localhost:5432 - no response
[Info] Waiting for the database to be ready...
localhost:5432 - no response
```

### To reproduce

N/A

### Sourcebot deployment information

Sourcebot version (e.g. v3.0.1): v3.0.2

msukkari:
The one time I've seen this happen is when postgres is unable to initialize on the initial startup. What is the setup you have with mounting the disk to the container?

gaeljw:
> What is the setup you have with mounting the disk to the container?

It's a regular volume mount through Kubernetes. Volume is NFS though. I've had issues in the past with NFS volumes and Postgres, so that could be it. But there's not much logs for me to investigate.

I'll give it locally on my laptop to troubleshoot

gaeljw:
This definitely look's like a permission issue.

On my laptop, I'm able to make it work but I see that the `postgres` folder is owned by another user:

```shell
ll .sourcebot   
total 16K
drwx------. 19   100069   100069 4,0K  7 avril 13:06 db
drwxr-xr-x.  2 gael gael 4,0K  7 avril 13:06 index
drwxr-xr-x.  2 gael gael 4,0K  7 avril 13:07 redis
drwxr-xr-x.  3 gael gael 4,0K  7 avril 13:06 repos
```

I know NFS is a bit crappy for such things. I've had a look at another setup where we have a NFS volume mount for a Postgres container and we've had to add a couple of configurations to make it work (`runAsUser` and `runAsGroup` on the postgres container with same uid than the owner of NFS files).

Thus I guess it's something related to our infra rather than an issue with Sourcebot.

[...]

Issue #168, opened 2025-01-17
Allow to customize the `.gitconfig`?


gaeljw:
Hello folks,

Would it make sense to allow to customize the `.gitconfig` used by Sourcebot through Sourcebot config? Or we say it's not Sourcebot responsibility and the user has to define the `.gitconfig` however needed depending on Sourcebot is installed (like mounting a volume to `/root/.gitconfig` if using the container/K8S approach).

I'm asking the question because I faced the need to configure the `.gitconfig` for a corner case today due to mounting the repos through a volume that didn't have the right file ownerships.

I was facing the following error:
```
[backend] | 2025-01-16T13:18:28.357Z error: [main] fatal: detected dubious ownership in repository at '/data/repos/gitlab.mycompany.net/mygroup/myrepo.git'
[backend] | To add an exception for this directory, call:
[backend] |
[backend] | 	git config --global --add safe.directory /data/repos/gitlab.mycompany.net/mygroup/myrepo.git
[backend] |
[backend] | Error: fatal: detected dubious ownership in repository at '/data/repos/gitlab.mycompany.net/mygroup/myrepo.git'
[backend] | To add an exception for this directory, call:
[backend] |
[backend] | 	git config --global --add safe.directory /data/repos/gitlab.mycompany.net/mygroup/myrepo.git
[backend] |
[backend] |     at Object.action (file:///app/node_modules/simple-git/dist/esm/index.js:4462:25)
[backend] |     at PluginStore.exec (file:///app/node_modules/simple-git/dist/esm/index.js:4501:25)
[backend] |     at file:///app/node_modules/simple-git/dist/esm/index.js:1369:43
[backend] |     at new Promise (<anonymous>)
[backend] |     at GitExecutorChain.handleTaskData (file:///app/node_modules/simple-git/dist/esm/index.js:1367:16)
[backend] |     at GitExecutorChain.<anonymous> (file:///app/node_modules/simple-git/dist/esm/index.js:1351:44)
[backend] |     at Generator.next (<anonymous>)
[backend] |     at fulfilled (file:///app/node_modules/simple-git/dist/esm/index.js:45:24)
[backend] |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

And I temporarily worked around it by adding a `.gitconfig` with following content in Sourcebot pod:
```
[safe]
	directory = *
```

[...]

PR #370, opened 2025-07-01
Add Sourcebot Helm Chart

aanogueira:
This PR adds an Helm Chart enabling the deployment of Sourcebot to Kubernetes.


msukkari:
This is awesome thanks Andre! Let us know if there's anything we can help with to get this PR out of drafts

aanogueira:
The chart it self should be good to go. In order to use it, the typical way is to provide a way to fetch the packaged chart.

This means hosting it on a registry. Alternatively, Github pages can be used for this - here is the documentation reference: https://helm.sh/docs/topics/chart_repository/

Added it as a Draft to discuss which approach would work best here.

gaeljw:
@aanogueira I haven't looked at the code yet (I'll be happy to give a review if you're interested ; we've built our own Helm chart internally and I was also considering contributing it to the community) but I was wondering if you were able to run Sourcebot in Kubernetes just as is with the embedded postgres server (starting from V3)? Asking because we faced errors related to the volume mount (https://github.com/sourcebot-dev/sourcebot/issues/263) and had to use a separate Postgres in the end.

kubernetes/chart/README.md
```diff
@@ -0,0 +1,92 @@
...
+| autoscaling | object | `{"enabled":false,"maxReplicas":3,"minReplicas":1,"targetCPUUtilizationPercentage":80,"targetMemoryUtilizationPercentage":80}` | Configure Horizontal Pod Autoscaler. |
+| autoscaling.enabled | bool | `false` | Enable or disable Horizontal Pod Autoscaler. |
+| autoscaling.maxReplicas | int | `3` | Maximum number of replicas. |
+| autoscaling.minReplicas | int | `1` | Minimum number of replicas. |
+| autoscaling.targetCPUUtilizationPercentage | int | `80` | Target CPU utilization percentage for autoscaling. |
+| autoscaling.targetMemoryUtilizationPercentage | int | `80` | Target memory utilization percentage for autoscaling. |
+| command | list | `[]` | Override the default command of the container. |
+| config | object | `{"$schema":"https://raw.githubusercontent.com/sourcebot-dev/sourcebot/main/schemas/v3/index.json","connections":{},"settings":{}}` | Configure Sourcebot-specific application settings. |
+| containerSecurityContext | object | `{}` | Set the container-level security context. |
+| database | object | `{}` | Configure the database secret. |
```
gaeljw: Reading this, it's unclear what is expected in the `database` key.

kubernetes/chart/README.md
```diff
@@ -0,0 +1,92 @@
...
+| envSecrets | list | `[]` | Set environment variables from Kubernetes secrets. |
+| envs | list | `[]` | Set additional environment variables. |
+| fullnameOverride | string | `""` | Override the full name of the chart. |
+| image | object | `{"pullPolicy":"Always","repository":"ghcr.io/sourcebot-dev/sourcebot","tag":"latest"}` | Configure the container image. |
+| image.pullPolicy | string | `"Always"` | Image pull policy. |
+| image.repository | string | `"ghcr.io/sourcebot-dev/sourcebot"` | Container image repository. |
+| image.tag | string | `"latest"` | Container image tag. |
```
gaeljw: Do we want to default to "latest"? It's usually a bad practice to use "latest" as you can't be sure of which version is deployed and it's not reproducible. I would stick a fixed version here by default, and let user change it optionally.

This also means that the release workflow should include bumping the default version of the image in the container and release a new version of the Helm chart.

kubernetes/chart/README.md
```diff
@@ -0,0 +1,92 @@
...
+| imagePullSecrets | list | `[]` | Configure image pull secrets for private registries. |
+| ingress | object | `{"annotations":{},"className":"","enabled":false,"hosts":[],"tls":[]}` | Configure ingress for Sourcebot. |
+| ingress.annotations | object | `{}` | Ingress annotations. |
+| ingress.className | string | `""` | Ingress class name. |
+| ingress.enabled | bool | `false` | Enable or disable ingress. |
+| ingress.hosts | list | `[]` | List of hostnames and paths for ingress rules. |
+| ingress.tls | list | `[]` | TLS settings for ingress. |
+| initContainers | list | `[]` | Configure init containers to run before the main container. |
+| license | object | `{}` | Configure the enterprise license key secret. |
```
gaeljw: Same here, unclear what kind of value is expected.

kubernetes/chart/README.md
```diff
@@ -0,0 +1,92 @@
...
+| podSecurityContext | object | `{}` | Set the pod-level security context. |
+| priorityClassName | string | `""` | Set the priority class name for pods. See: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/ |
+| readinessProbe | object | `{"failureThreshold":5,"httpGet":{"path":"/","port":"http"},"initialDelaySeconds":10,"periodSeconds":10}` | Readiness probe to check if the container is ready to serve traffic. |
+| readinessProbe.failureThreshold | int | `5` | Number of consecutive failures before marking the container as not ready. |
+| readinessProbe.httpGet | object | `{"path":"/","port":"http"}` | Http GET request to check if the container is ready. |
+| readinessProbe.httpGet.path | string | `"/"` | Path to check. |
+| readinessProbe.httpGet.port | string | `"http"` | Port to check. |
+| readinessProbe.initialDelaySeconds | int | `10` | Initial delay before the first probe. |
+| readinessProbe.periodSeconds | int | `10` | Frequency of the probe. |
+| redis | object | `{}` | Configure the Redis secret. |
```
gaeljw: Same comment here :)

kubernetes/chart/README.md
```diff
@@ -0,0 +1,92 @@
...
+| serviceAccount.name | string | `""` | Use an existing ServiceAccount (if set). |
+| startupProbe | object | `{"failureThreshold":30,"httpGet":{"path":"/","port":"http"},"periodSeconds":30}` | Startup probe to check if the container has started successfully. |
+| startupProbe.failureThreshold | int | `30` | Number of seconds to wait before starting the probe. |
+| startupProbe.httpGet | object | `{"path":"/","port":"http"}` | Http GET request to check if the container has started. |
+| startupProbe.httpGet.path | string | `"/"` | Path to check. |
+| startupProbe.httpGet.port | string | `"http"` | Port to check. |
+| startupProbe.periodSeconds | int | `30` | Initial delay before the first probe. |
+| tolerations | list | `[]` | Set tolerations for pod scheduling. See: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ |
+| volumeMounts | list | `[]` | Define volume mounts for the container. |
+| volumes | list | `[]` | Define additional volumes. |
```
gaeljw: We should probably give some guidance as to what to configure for volumes (especially path expected by Sourcebot). Unless I misread, if nothing is specified, there's no persistence in the current version of the chart.

kubernetes/chart/README.md
aanogueira: It's in the `values.yaml`, but added to the generated docs as well for clarity.

kubernetes/chart/README.md
aanogueira: This was me testing it out, and forgot to remove it from the default values.

Added the specific version to the `appVersion` section of the `Chart.yaml` metadata.

kubernetes/chart/README.md
aanogueira: It's in the `values.yaml`, but added to the generated docs as well for clarity.

kubernetes/chart/README.md
aanogueira: It's in the `values.yaml`, but added to the generated docs as well for clarity.

kubernetes/chart/README.md
aanogueira: The `volumes`/`volumeMounts` follow the default Kubernetes specs.

As for persistence, I added support on the new commit to the `storage` option to support the usage of the internal DB, mounting to the pod under `/data` path.

aanogueira:
> @aanogueira I haven't looked at the code yet (I'll be happy to give a review if you're interested ; we've built our own Helm chart internally and I was also considering contributing it to the community) but I was wondering if you were able to run Sourcebot in Kubernetes just as is with the embedded postgres server (starting from V3)? Asking because we faced errors related to the volume mount (#263) and had to use a separate Postgres in the end.

No, there was no persistence. For live deployments, I was not assuming the usage of the internal DB, but indeed there might be some scenario where it could be useful. For those cases, we could deploy a StatefulSet instead of Deployment.

Added a new block (storage) to add persistence storage, changing the deployable from a Deployment to a StatefulSet.

brendan-kellam:
Hey @aanogueira thanks for the PR!! Apologies for the delay on our end - has been hectic shipping Ask SB 🤠 Planning on getting to this this week. I've only _briefly_ used helm charts before, so I'll probably ask you a bunch of dumb questions as I ramp up on things 😄 On my side, I will get a minikube instance running locally and test out the chart

komapa:
I am following this PR mostly to see how we are supposed to scale the sourcebot deployment past as single pod. Sure, we can run StatefulSet and also an external database but what about the storage itself? Most storage drivers do not allow many pods writing to the same PVC but let's say we can, how does Sourcebot handle this correctly?

brendan-kellam:
> I am following this PR mostly to see how we are supposed to scale the sourcebot deployment past as single pod. Sure, we can run StatefulSet and also an external database but what about the storage itself? Most storage drivers do not allow many pods writing to the same PVC but let's say we can, how does Sourcebot handle this correctly?

Hey @komapa - not certain on what the approach will be here, but speaking to multiple-writers: I think the only thing that needs to **write** to the volume are the "workers" (i.e., `packages/backend`). We are already using a redis queue to process index jobs, so presumably, we should be able to scale to multiple worker replicas w/ the redis queue preventing race conditions on job processing.

Does this approach make sense?

*Edit: working on a RFC here: #439

brendan-kellam:
> The chart it self should be good to go. In order to use it, the typical way is to provide a way to fetch the packaged chart.
> 
> This means hosting it on a registry. Alternatively, Github pages can be used for this - here is the documentation reference: https://helm.sh/docs/topics/chart_repository/
> 
> Added it as a Draft to discuss which approach would work best here.

I think I would prefer Github pages since we can keep everything under one umbrella. I was reading up on it and it seems the best option for that is to host the chart from a separate repo, so I created this one: https://github.com/sourcebot-dev/sourcebot-k8s

kubernetes/chart/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| additionalLabels | object | `{}` | Add extra labels to all resources. |
+| affinity | object | `{}` | Set affinity rules for pod scheduling. Defaults to soft anti-affinity if not set. See: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/ |
+| args | list | `[]` | Override the default arguments of the container. |
+| autoscaling | object | `{"enabled":false,"maxReplicas":3,"minReplicas":1,"targetCPUUtilizationPercentage":80,"targetMemoryUtilizationPercentage":80}` | Configure Horizontal Pod Autoscaler. |
+| autoscaling.enabled | bool | `false` | Enable or disable Horizontal Pod Autoscaler. |
```
brendan-kellam: I'm not sure if we want to include autoscaling settings at the moment since Sourcebot isn't setup with horizontal scale in mind atm

kubernetes/chart/templates/deloyable.yaml
```diff
@@ -0,0 +1,159 @@
+---
+apiVersion: apps/v1
+{{- if $.Values.storage }}
+kind: StatefulSet
```
brendan-kellam: > > @aanogueira I haven't looked at the code yet (I'll be happy to give a review if you're interested ; we've built our own Helm chart internally and I was also considering contributing it to the community) but I was wondering if you were able to run Sourcebot in Kubernetes just as is with the embedded postgres server (starting from V3)? Asking because we faced errors related to the volume mount (#263) and had to use a separate Postgres in the end.
> 
> No, there was no persistence. For live deployments, I was not assuming the usage of the internal DB, but indeed there might be some scenario where it could be useful. For those cases, we could deploy a StatefulSet instead of Deployment.
> 
> Added a new block (storage) to add persistence storage, changing the deployable from a Deployment to a StatefulSet.

Just to double check: could we use just a Deployment here if we did not support the internal DB case?

I think the direction I would like to move in is moving upstream dependencies (postgres, redis) out of our container and instead have them specified as dependencies. For a local deployments, we could use docker-compose with [depends_on](https://docs.docker.com/reference/compose-file/services/#depends_on). For the helm chart, we can use [dependencies](https://helm.sh/docs/helm/helm_dependency/). We don't need to address this in this PR, but something I've been  thinking about.

kubernetes/chart/templates/config.yaml
brendan-kellam, line 1: TIL this is a thing in k8s 🤠

brendan-kellam (approved):
I'm good with merging this. Thanks @gaeljw for also reviewing!! Once this is in, I can figure out deploying it on GitHub pages, and then write up some quick docs.

I think next up w.r.t., deployment stuff: I'm going to get started on a RFC in #439  with some of my thoughts on packaging our individual services into separate containers (i.e., `ghcr.io/sourcebot-dev/(webapp|worker|zoekt)`), as well as some thoughts on how we can support multiple replicas per service. Please feel free to add your own thoughts!

gaeljw:
For the record, the setup we currently use at my company is the following:
- Separate Postgres (could be a dependency in the Helm chart, it has to be modifiable so that an external Postgres can be used though)
- a `Deployment` for Sourcebot, with a single replica
  - injecting environment variables (like `GITLAB_TOKEN`, `DATABASE_URL`, `OPENAI_API_KEY`) through a reference to a `Secret`, I think this would be nice if not mandatory to have as well (the secret name should be configurable then) ; typically we populate the `Secret` using external-secrets-provider
  - a volume for `/data` (data cache) mounted from persistent storage (that works fine because we have a single replica, otherwise all replicas would share the same data cache and this would cause issues from what I read above?)
  - a volume for `/config` (config file) mounted from a `ConfigMap`
  - a volume for `/root/.gitconfig` mounted from a `ConfigMap` ; this one seem to be specific to us? (https://github.com/sourcebot-dev/sourcebot/issues/168)
- a `ConfigMap` for config file 
- a `ConfigMap` for the git config
- a `Service`
- an `Ingress`

kubernetes/chart/templates/deloyable.yaml
gaeljw: In my understanding, the `Deployment` only works if there's a single replica of Sourcebot. Otherwise the same volume (cache) is shared across the replicas (all replicas read/write to it) and this will cause conflicts/issues in Sourcebot, isn't it?

(That being said, we do use a `Deployment` with a single replica at my company, to keep things simple)

gaeljw:
> it seems the best option for that is to host the chart from a separate repo, so I created this one: https://github.com/sourcebot-dev/sourcebot-k8s

@brendan-kellam I believe naming the repo `sourcebot-helm-chart` would make more sense. That's what most projects do. `k8s` is a bit broad in terms of naming. Unless you plan to include other things than the Helm chart in this repo, of course.

kubernetes/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,159 @@
...
+          {{- end }}
+          image: {{ include "sourcebot.image" $ }}
+          imagePullPolicy: {{ $.Values.image.pullPolicy }}
+          {{- with $.Values.command }}
+          command: {{ toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with $.Values.args }}
+          args: {{ toYaml . | nindent 12 }}
+          {{- end }}
+          env:
```
gaeljw, line 52: Shouldn't we set the `AUTH_URL` env property automatically from the `Ingress` host? Or defaults to it and allow overriding?

kubernetes/chart/README.md
aanogueira: Fine with me. Made it a bit generic thus the possibility to scale it. But teh charts needs to be adapted to the app it self, and if additional replicas is not advised, I'll adapt it. 😄

kubernetes/chart/templates/deloyable.yaml
aanogueira: We could include Postgres as an additional chart dependency. Personally I prefer deploying the database separately.

This approach removes the need for persistence (StatefulSet) within the Sourcebot deployment. The chart provides an optional Postgres setup to make it easy for quick trials, but in the intended setup the database should be externally managed.

kubernetes/chart/templates/deployment.yaml
aanogueira: Yeah, it can definitely be inferred from it.

aanogueira:
> I'm good with merging this. Thanks @gaeljw for also reviewing!! Once this is in, I can figure out deploying it on GitHub pages, and then write up some quick docs.
> 
> I think next up w.r.t., deployment stuff: I'm going to get started on a RFC in #439 with some of my thoughts on packaging our individual services into separate containers (i.e., `ghcr.io/sourcebot-dev/(webapp|worker|zoekt)`), as well as some thoughts on how we can support multiple replicas per service. Please feel free to add your own thoughts!

I can help with that if needed! 😄

kubernetes/chart/README.md
brendan-kellam: Yea I would say let's remove it for now and we can add it back when we officially support it

kubernetes/chart/templates/deloyable.yaml
pjbgf: > We could include Postgres as an additional chart dependency. Personally I prefer deploying the database separately.

+1. That better aligns with HA enablement and production deployment models.

It would be nice if there was a container image that only contained the application, as opposed to having redis and postgres in it. But that is largely orthogonal to this change. (just saw that this is already being discussed elsewhere - so please disregard)

kubernetes/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,159 @@
+---
```
pjbgf, line 1: Would it be OK to rename this file to `deployment.yaml` instead?

kubernetes/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,159 @@
...
+      {{- with $.Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "sourcebot.serviceAccountName" $ }}
+      {{- with $.Values.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
```
pjbgf, line 38: The `values.yaml` defines `initContainers`, but that is not being used here. Do you mind adding it here please? That would enable some use cases as the x-ref'd.

kubernetes/chart/values.yaml
```diff
@@ -0,0 +1,279 @@
...
+# -- Set environment variables from Kubernetes secrets.
+envSecrets: []
+# - secretName: sourcebot-github-token
+#   secretKey: token
+#   envName: GITHUB_TOKEN
+
+# -- Set additional environment variables.
+envs: []
+# - name: FOO
+#   value: bar
```
pjbgf: Maybe we could add examples on how to disable telemetry:
```suggestion
# # Disable sourcebot telemetry.
# - name: SOURCEBOT_TELEMETRY_DISABLED
#   value: "1"
# # Disable Prisma checkpoint.
# - name: CHECKPOINT_DISABLE
#   value: "1"
```

pjbgf (commented):
@aanogueira really good stuff, thanks for putting this together. 🙇 

Added a few nits but overall LGTM.

kubernetes/chart/values.yaml
aanogueira: Yes, this makes a lot more sense for the application!

kubernetes/chart/templates/deployment.yaml
aanogueira: Missed this one... will map it.

kubernetes/chart/templates/deployment.yaml
aanogueira: Yes, without the STS I will rename it accordingly.

PR #372, opened 2025-07-02
Improved Gitlab All in case of many repositories

abraverm:
When there are many repositories on Giltab, doing actions in one hit would fail to API timeout.
This is a quick workaround/hack by using batches. 

Also when there a lot of repositories, just getting the list of them might be too much.

Ideally, I would like Sourcebot to add a repository as it finds it instead of handling the entire connection sync. Otherwise this process blocks other connections from syncing, and if it fails/breaks, then it has to start over.

Note: prisma also failing when there are too many connections, not sure what is the magical number is (sometimes 1000 resulted with errors too):
```
Invalid `prisma.userToOrg.findUnique()` invocation: Too many database connections opened: FATAL: remaining connection slots are reserved for roles with the SUPERUSER attribute Prisma Accelerate has built-in connection pooling to prevent such errors: https:/ / pris. ly/ client/ error-accelerate
( https://pris.ly/client/error-accelerate ) at Vn.handleRequestError
(/nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/node_modules/@prisma/client/runtime/library.js:121:7339) at Vn.handleAndLogRequestError
(/nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/node_modules/@prisma/client/runtime/library.js:121:6663) at Vn.request
(/nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/node_modules/@prisma/client/runtime/library.js:121:6370) at async l
(/nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/node_modules/@prisma/client/runtime/library.js:130:9617) at async el (/ nix/ store/ sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/ packages/ web/. next/ server/ chunks/ 616. js:1:39994 (
http://nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/packages/web/.next/server/chunks/616.js:1:39994
) ) at async ei (/ nix/ store/ sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/ packages/ web/. next/ server/ chunks/ 616. js:1:38467 ( http://nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/packages/web/.next/server/chunks/616.js:1:38467
) ) at async j (/ nix/ store/ sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/ packages/ web/. next/ server/ app/ [domain]/ page. js:1:4267 ( http://nix/store/sbwsr0wj0665xvgzg2iqim0c2qdynqkw-sourcebot/packages/web/.next/server/app/[domain]/page.js:1:4267
) ) { code: 'P2037',
clientVersion: '6.2.1',
meta: {
modelName: 'UserToOrg',
message: 'FATAL: remaining connection slots are reserved for roles with the SUPERUSER attribute'
}
}
```

### DIFF
--- a/packages/backend/src/gitlab.ts
+++ b/packages/backend/src/gitlab.ts
@@ -7,6 +7,7 @@ import { PrismaClient } from "@sourcebot/db";
 import { processPromiseResults, throwIfAnyFailed } from "./connectionUtils.js";
 import * as Sentry from "@sentry/node";
 import { env } from "./env.js";
+import { log } from "console";
 
 const logger = createLogger('gitlab');
 export const GITLAB_CLOUD_HOSTNAME = "gitlab.com";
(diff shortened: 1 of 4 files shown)


packages/backend/src/gitlab.ts
```diff
@@ -65,76 +78,96 @@ export const getGitLabReposFromConfig = async (config: GitlabConnectionConfig, o
...
+                }
+            }));
+            allResults.push(...batchResults);
+        }
+        const { validItems: validRepos, notFoundItems: notFoundOrgs } = processPromiseResults(allResults);
         allRepos = allRepos.concat(validRepos);
         notFound.orgs = notFoundOrgs;
+        logger.debug(`Found ${validRepos.length} valid repositories in groups.`);
+        logger.debug(`Not found groups: ${notFoundOrgs.join(', ')}`);
+        logger.debug(`These repositories will be downloaded: ${allRepos.map(repo => repo.path_with_namespace).join('\n')}`);
```
msukkari, line 126: nit: `synced` not `downloaded` to be consistent with language elsewhere

msukkari (approved):

msukkari:
Thanks for the change! I assume you've already tested this with larger configs? Would love to know if you've noticed a difference

abraverm:
The change is necessary for the sync to work at all, otherwise it fails on timeout. I don't think you should merge the code as is or continue with the batch direction. I think this part should be refactored so every batch of found repository would be processed (filterd/checked) and added to the database and to the sync pipeline. Because when you have a lot of repositories, this step takes a very long time (10 minutes?) and if something fails it might need to start again. It would be nice to see a stream of repositories being detected and processed. Also "100" batch is right for me but maybe not for another instance, another configuration like that is possible but magic numbers are annoying.

abraverm:
The too many connections persists outside the sync flow. I managed to overcome it with PgBouncer: https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/databases-connections/pgbouncer

PR #381, opened 2025-07-16
[bug] onboarding: fix infinite loop when using GCP IAP

drew-u410:
* If GCP IAP Auth is used, you will enter an infinite loop when trying to auth due to updated onboarding flow.
* This does not show the onboarding flow if these variables are set.

Repro:
* Set the `AUTH_EE_GCP_IAP_ENABLED` and `AUTH_EE_GCP_IAP_AUDIENCE` and try to login.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue causing an onboarding infinite loop when GCP IAP Authentication is enabled.

* **Documentation**
  * Updated the changelog with details about the onboarding loop fix under a new "Unreleased" section.


msukkari:
Thanks for highlighting this @drew-u410! 

I had added the following logic to the onboarding page to support the IAP bridge if the user isn't onboarded, but admittingly I didn't fully test this path: 
https://github.com/sourcebot-dev/sourcebot/blob/ee90edc46d823c0af2686afe6211a615df0f037f/packages/web/src/app/onboard/page.tsx#L81

I'm trying to figure out the right approach here, because ideally we wouldn't have the onboarding flow be complete skipped when an IAP bridge is used.

Was your org not onboarded when you're hitting this infinite loop? Existing orgs should have had their `isOnboarded` flag be set to true

msukkari:
If you could describe what the "infinite loop" while authing looked like it would help to figure out the underlying issue here

drew-u410:
> Was your org not onboarded when you're hitting this infinite loop? Existing orgs should have had their `isOnboarded` flag be set to true

No idea, it just loops trying to load / auth.

Just configure with these two env vars and try to login (you can't).

> I'm trying to figure out the right approach here, because ideally we wouldn't have the onboarding flow be complete skipped when an IAP bridge is used.

I think this is probably? okay.  If you are setting these... you are probably way past onboarding.

msukkari:
Sounds good, I will setup my GCP IAP environment and try to repro this as soon as I get the chance

msukkari:
Going to merge this in for now so we don't block Ask Sourcebot release. Will revisit when I get the chance to find a longer term fix

PR #385, opened 2025-07-18
Add anonymous access option to core

msukkari:
- Adds an anonymous access option to core, allowing a Sourcebot instance to be accessed without needing to log in. 
  - Note that deployments with an enterprise license cannot enable this option. If you're using an IAP please configure an [IAP bridge](http://localhost:3001/docs/configuration/auth/providers#enterprise-authentication-providers). If your IAP isn't supported please [reach out](https://www.sourcebot.dev/contact) 

- Adds a toggle to control this in onboarding and in a new `Access` page in the settings
- Also adds the `FORCE_ENABLE_ANONYMOUS_ACCESS` env var to force enable anonymous access

image


schemas/v3/index.json
```diff
@@ -61,11 +61,6 @@
                     "type": "number",
                     "description": "The timeout (in milliseconds) for a repo indexing to timeout. Defaults to 2 hours.",
                     "minimum": 1
-                },
-                "enablePublicAccess": {
```
brendan-kellam, line 65: This is a breaking change. Maybe we can mark this field as deprecated and continue to support it until we release a Schema v4

packages/web/src/initialize.ts
```diff
@@ -105,23 +107,27 @@ const syncConnections = async (connections?: { [key: string]: ConnectionConfig }
 const syncDeclarativeConfig = async (configPath: string) => {
     const config = await loadConfig(configPath);
 
-    const hasPublicAccessEntitlement = hasEntitlement("public-access");
```
brendan-kellam, line 108: Even though the setting is marked deprecated, we will need to still support it, otherwise this is a breaking change. Maybe `const isAnonymousAccessEnabled = env.FORCE_ENABLE_ANONYMOUS_ACCESS || config.settings?.enablePublicAccess`

docs/docs/configuration/auth/access-settings.mdx
```diff
@@ -1,13 +1,21 @@
...
-There are various ways to configure how members can join a Sourcebot deployment. 
+There are various settings to control how users access your Sourcebot deployment.
 
-## Member Approval 
+# Anonymous access
 
-**By default, Sourcebot requires new members to be approved by the owner of the deployment**. This section explains how approvals work and how
+<Note>Anonymous access cannot be enabled if you have an enterprise license. If you have any questions about this restriction [reach out to us](https://www.sourcebot.dev/contact).</Note>
+
+By default, your Sourcebot deployment is gated with a login page. If you'd like users to access the deployment anonymously, you can enable anonymous access.
```
brendan-kellam, line 12: Maybe we could add a little bit more detail here about what an anonymous user can and cannot do? And how it ties into the role a user can take on in the organization?

Also just noticed that the roles docs are a bit out of date and don't list the guest role.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Fixed typos in UI, docs, code [#369](https://github.com/sourcebot-dev/sourcebot/pull/369)
+- Add anonymous access option to core [#385](https://github.com/sourcebot-dev/sourcebot/pull/385)
```
brendan-kellam, line 13: Could you add a entry that the `enablePublicAccess` config setting has been deprecated?

packages/web/src/actions.ts
```diff
@@ -62,13 +63,13 @@ export const sew = async <T>(fn: () => Promise<T>): Promise<T | ServiceError> =>
     }
 }
 
-export const withAuth = async <T>(fn: (userId: string, apiKeyHash: string | undefined) => Promise<T>, allowSingleTenantUnauthedAccess: boolean = false, apiKey: ApiKeyPayload | undefined = undefined) => {
+export const withAuth = async <T>(fn: (userId: string, apiKeyHash: string | undefined) => Promise<T>, allowAnonymousAccess: boolean = false, apiKey: ApiKeyPayload | undefined = undefined) => {
```
brendan-kellam, line 66: Not asking for us to do this in this PR, but I feel like this allowAnonymous flag is a bit of a code-smell. I'm pretty sure that in all scenarios where this is set to `true`, the minimum organization role will be `GUEST` anyways. If a organization has anonymous access enabled, then the minimum role remains at `GUEST`. If not, then the minimum role for all actions is raised to `MEMBER`.

packages/web/src/types.ts
```diff
@@ -1,7 +1,13 @@
+import { Org } from "@sourcebot/db";
 import { z } from "zod";
 
 export const orgMetadataSchema = z.object({
-    publicAccessEnabled: z.boolean().optional(),
+    anonymousAccessEnabled: z.boolean().optional(),
```
brendan-kellam, line 4: This looks like it will break back-compat - `publicAccessEnabled` will still be set in the DB's metadata, so when we go to parse the JSON blob it will fail. It would be worth testing the upgrade path

packages/web/src/types.ts
```diff
@@ -1,7 +1,13 @@
...
 })
 
-export type OrgMetadata = z.infer<typeof orgMetadataSchema>;
\ No newline at end of file
+export type OrgMetadata = z.infer<typeof orgMetadataSchema>;
+
+export const getOrgMetadata = (org: Org): OrgMetadata | null => {
```
brendan-kellam: nit: maybe better to be in a utils file?

packages/web/src/actions.ts
brendan-kellam: Thinking about it more, we could probably eliminate the need for the concept of our sentinel `guestUser` entirely, since everywhere where a user is _actually_ required (i.e., besides verifying they are a member of the organization), the role would be `MEMBER` or above.

packages/web/src/app/[domain]/settings/access/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+        <div className="flex flex-col gap-6">
+            <div>
+                <h3 className="text-lg font-medium">Access Control</h3>
+                <p className="text-sm text-muted-foreground">Configure how users can access your Sourcebot deployment.{" "}
+                    <a
+                        href="https://docs.sourcebot.dev/docs/configuration/auth/access-settings"
+                        target="_blank"
+                        rel="noopener"
+                        className="underline text-[var(--primary)] hover:text-[var(--primary)]/80 transition-colors"
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/settings/access/page.tsx]: minor nit: these tailwind classes look weird. They can just be `text-primary`. Maybe we want to add a cursor / claude code rule for this?

Also heads up: right now our tailwind config is a bit busted where opacity settings (e.g., `/80`) don't work (last time I checked)

Comment (brendan-kellam) [reply packages/web/src/app/[domain]/settings/access/page.tsx]: See: https://v3.tailwindcss.com/docs/customizing-colors#using-css-variables
#387

packages/web/src/initialize.ts
```diff
@@ -192,12 +198,21 @@ const initSingleTenancy = async () => {
...
         if (isServiceError(res)) {
             throw new ServiceErrorException(res);
         }
+    } else {
+        // If anonymous access entitlement is not enabled, set the flag to false in the org on init
+        const org = await getOrgFromDomain(SINGLE_TENANT_ORG_DOMAIN);
+        if (org) {
+            await prisma.org.update({
+                where: { id: org.id },
+                data: { metadata: { anonymousAccessEnabled: false } },
```
brendan-kellam: nit: we only have a single field in `metadata` so it doesn't matter rn, but if in the future we had more fields, would this override them? Do we need a `...` operator here?

packages/web/src/initialize.ts
msukkari: done

`const forceEnableAnonymousAccess = config.settings?.enablePublicAccess ?? env.FORCE_ENABLE_ANONYMOUS_ACCESS === 'true';`

docs/docs/configuration/auth/access-settings.mdx
msukkari: Added `Guest` role in docs and mention that in anonymous access

packages/web/src/actions.ts
msukkari: Yeah it definitely isn't ideal, but the reason this is needed rn is how we're chaining `withAuth` and `withOrgMembership`. The `withOrgMembership` expects `withAuth` to pass, so we're required to create the sentinel `Guest` user because it expects a `userId`. We need to pass in the `allowAnonymous` flag to know whether or not to return a `Guest` user or nothing

brendan-kellam: Right. We should probably just combine `withOrgMembership` and `withAuth` into a single middleware function that handles both the authentication and authorization checks.

packages/web/src/types.ts
msukkari: confirmed doesn't break back compat

packages/web/src/initialize.ts
msukkari: not a nit, nice catch

PR #420, opened 2025-08-01
docs: fix link in deployment guide next steps

nullun:


msukkari (approved):
oops, thanks for the catch!

PR #447, opened 2025-08-07
fix(ask_sb): Fix long generation times on first message in thread

brendan-kellam:
This PR fixes (or at least reduces) a issue we've seen where the first message in a thread takes longer to generate. I believe this was caused because the "generate title" step was blocking the actual chat stream. I observed ~10-15s generation times for gemini-2.5-pro.

This PR does not address making the title generation faster. Instead, I've made it so the title generation happens asynchronously from the chat stream s.t., it won't block it.


packages/web/src/features/chat/components/chatThread/chatThread.tsx
```diff
@@ -118,8 +120,38 @@ export const ChatThread = ({
...
+                    message: message.parts[0].text,
+                },
+                domain
+            ).then((response) => {
+                if (isServiceError(response)) {
+                    toast({
+                        description: `❌ Failed to generate chat name. Reason: ${response.message}`,
+                        variant: "destructive",
+                    });
+                }
```
brendan-kellam, line 143: Note: there is a race condition atm where if the stream (somehow) finishes before the chat title is generated, then once the new title completes it will not appear. This is because there won't be a router.refresh event.

msukkari: might be worth adding a TODO comment?

msukkari (approved):

packages/web/src/features/chat/components/chatThread/chatThread.tsx
brendan-kellam: Just pushed a fix for this

PR #480, opened 2025-08-26
feat(deployment): Basic docker-compose file

brendan-kellam:
Adds a basic docker-compose file to simplify the local deployment story of Sourcebot. Also updates the deployment docs to use the docker-compose file.

## Summary by CodeRabbit

- New Features
  - Added a Docker Compose service for running the app locally with a single command.
  - Exposes the web UI on port 3000, supports persistent storage, auto-restart, and configuration via environment variables.

- Documentation
  - Rewrote the deployment guide to a Docker Compose–based workflow with a step-by-step walkthrough.
  - Added a curl command to fetch the compose file and updated requirements.
  - Replaced docker run instructions with docker compose up -d.
  - Improved onboarding link and added references to configuration docs and support.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
 contact_links:
   - name: 👾 Discord
-    url: https://discord.gg/f4Cbf3HT
+    url: https://discord.gg/GbXMEM5H
     about: Something else? Join the Discord!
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,62 @@
+services:
```
msukkari, line 1: this might be a good opportunity to remove the postgres and redis dependencies from the core sourcebot image? or are you planning on doing that later since its a breaking change

msukkari (approved):

docker-compose.yml
brendan-kellam: That would be a breaking change, so something we will need to do in the next major release

PR #497, opened 2025-09-08
fix(backend): Explicitly set refspec on fetch

### LINKED ISSUE #492: Git repo state seems to be corrupt?
[backend] | 2025-09-05T20:44:17.357Z error: [repo-manager] Failed to sync repository [gerrit.hostname.com/MyRepo](http://gerrit.hostname.com/MyRepo) (id: 8) after 3 attempts. Error: Error: Failed to fetch repository: /data/.sourcebot/repos/8. Reason: fatal: couldn't find remote ref HEAD
[backend] |

I updated Sourcebot from v4.6.3 -> v4.6.6, running in k8s. terminationGracePeriod for my deployment was set to 15s (meaning k8s would send SIGTERM to Sourcebot process, wait 15 seconds, hope Sourcebot gracefully shut down, then send SIGKILL after 15 seconds regardless). Perhaps this was too low. 

In either case, after Sourcebot came up with v4.6.6 image, certain repos are now failing to sync/index, and retrying doesn't help. All fail with the same error signature (above). 

Not sure how to debug this / bandaid fix these failing repos. Redeploying SB from scratch will certainly fix things but that's sort of undesirable when there are 5000+ repos indexed already.

brendan-kellam:
For certain repos, re-indexing would fail when calling `git fetch` with error message `fatal: couldn't find remote ref HEAD`. The theory is that while we did have a [refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec) for `remote.origin.fetch`, it was not being resolved since we were fetching against the remote url and **not** `origin`. The solution is to specify the refspec inline with the fetch command.

Fixes #492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed "couldn't find remote ref HEAD" errors when re-indexing certain repositories. [#497](https://github.com/sourcebot-dev/sourcebot/pull/497)
+
 ### Changed
 - Disable page scroll when using arrow keys on search suggestions box. [#493](https://github.com/sourcebot-dev/sourcebot/pull/493)
 


brendan-kellam:
@musabshak do you mind pasting the error message you were seeing during fetch so I can include it in the PR description? Nvm it's in the issue 😅

Issue #295, opened 2025-05-07
[bug] Bitbucket cloud failed to get repos


WendelBartzUbots:
### Describe the bug

[backend] | 2025-05-07T14:33:55.882Z error: [Bitbucket] Failed to get repos for workspace *****: Error: Failed to fetch projects for workspace *****: {"type":"error","error":{"message":"Resource not found","detail":"There is no API hosted at this URL.\n\nFor information about our API's, please refer to the documentation at: https://developer.atlassian.com/bitbucket/api/2/reference/"}}

[backend] | 2025-05-07T14:33:55.883Z error: [ConnectionManager] Failed to compile repo data for connection 1 (starter-connection): Error: Failed to fetch projects for workspace *****: {"type":"error","error":{"message":"Resource not found","detail":"There is no API hosted at this URL.\n\nFor information about our API's, please refer to the documentation at: https://developer.atlassian.com/bitbucket/api/2/reference/"}}

### To reproduce

This is my config:

```
{
    "$schema": "https://raw.githubusercontent.com/sourcebot-dev/sourcebot/main/schemas/v3/index.json",
    "connections": {
        "starter-connection": {
            "type": "bitbucket",
            "deploymentType": "cloud",
            "user": "*****",
            "token": {
                "env": "BITBUCKET_TOKEN"
            },
            "workspaces": [
              "*****"
            ]
        }
    }
}
```

### Sourcebot deployment information

Sourcebot version (e.g. v3.0.1): ghcr.io/sourcebot-dev/sourcebot:main

msukkari:
Hey can you try and see if this same error happens if you clone a public bitbucket cloud repo with no auth? It might be an issue with how auth has been implemeneted for Bitbucket cloud

bardock:
I was able to resolve this issue in #502 

@WendelBartzUbots, could you please verify that this works with your configuration?
I noticed that authentication with access tokens isn't working for me, but app passwords work fine.

@msukkari, I'd appreciate a merge soon so we can start using this tool.
Please let me know if there's anything that needs to be improved or corrected.

WendelBartzUbots:
@bardock I ran tests locally and noticed that page 1 returns the repositories, but when it tries page 2, it comes back with an empty response, which causes the following error: 2025-09-10T15:06:57.880Z error: [bitbucket] Failed to get repos for workspace ****: TypeError: Cannot read properties of undefined (reading 'values')

brendan-kellam:
[...]

PR #502, opened 2025-09-09
fix(bitbucket): Bitbucket Cloud pagination not working beyond first page

bardock:
# Fix: Bitbucket Cloud pagination not working beyond first page

## Problem

Issue: #295 
When fetching repositories from Bitbucket Cloud workspaces with multiple pages of results, Sourcebot fails on the second page with the following error:

```
[dev:backend] 2025-09-09T19:11:45.876Z error: [bitbucket] Failed to get repos for workspace emilabs: Error: Failed to fetch projects for workspace emilabs: {"type":"error","error":{"message":"Resource not found","detail":"There is no API hosted at this URL.\n\nFor information about our API's, please refer to the documentation at: https://developer.atlassian.com/bitbucket/api/2/reference/"}}
```

This is caused by the Bitbucket Cloud API returning full URLs (e.g., `https://api.bitbucket.org/2.0/repositories/workspace?page=2`) in pagination `next` responses, but `openapi-fetch` expects separate path and query parameters. The current implementation was passing the full URL directly to the API client, which doesn't recognize the complete URL format.

## Solution

This PR implements proper URL parsing for Bitbucket Cloud pagination to separate paths and query parameters. The implementation includes:

### 1. **URL Parsing Function**

* Created `parseUrl()` function that properly extracts pathname and query parameters from full URLs
* Uses the native `URL` constructor for robust and performant parsing

### 2. **Updated Pagination Handlers**

* Updated `cloudGetReposForWorkspace()` to use parsed URL components
* Updated `cloudGetReposForProjects()` to use parsed URL components  
* Both functions now correctly pass `path` and `query` parameters separately to `openapi-fetch`

## Testing

* Manually tested with Bitbucket Cloud workspaces containing multiple pages of repositories
* Verified that pagination now works correctly beyond the first page
* Confirmed that single-page results continue to work as expected
* All existing tests are still passing

## Breaking Changes

None. The changes are backward compatible and only affect the internal pagination URL handling.

## Notes

* This fix specifically addresses Bitbucket Cloud pagination - Bitbucket Server has another implementation for pagination
* Debug logging includes the parsed path and query for troubleshooting pagination issues

### DIFF
--- a/packages/backend/src/bitbucket.ts
+++ b/packages/backend/src/bitbucket.ts
@@ -170,21 +170,33 @@ const getPaginatedCloud = async <T>(
     }
     return results;
 }
-   
+
+/**
+ * Parse the url into a path and query parameters to be used with the api client (openapi-fetch)
+ */
+function parseUrl(url: string, baseUrl: string): { path: string; query: Record<string, string>; } {
+    const fullUrl = new URL(url);
+    const path = fullUrl.pathname;
+    const query = Object.fromEntries(fullUrl.searchParams);
+    logger.debug(`Parsed url ${url} into path ${path} and query ${JSON.stringify(query)}`);
+    return { path, query };
+}
+
 
 async function cloudGetReposForWorkspace(client: BitbucketClient, workspaces: string[]): Promise<{validRepos: CloudRepository[], notFoundWorkspaces: string[]}> {
     const results = await Promise.allSettled(workspaces.map(async (workspace) => {
         try {
             logger.debug(`Fetching all repos for workspace ${workspace}...`);
 
-            const path = `/repositories/${workspace}` as CloudGetRequestPath;
             const { durationMs, data } = await measure(async () => {
-                const fetchFn = () => getPaginatedCloud<CloudRepository>(path, async (url) => {
-                    const response = await client.apiClient.GET(url, {
+                const fetchFn = () => getPaginatedCloud<CloudRepository>(`/repositories/${workspace}` as CloudGetRequestPath, async (url) => {
+                    const { path, query } = parseUrl(url, client.baseUrl);
+                    const response = await client.apiClient.GET(path, {
                         params: {
                             path: {
                                 workspace,
-                            }
+                            },
+                            query: query,
                         }
                     });
                     const { data, error } = response;
@@ -238,11 +250,15 @@ async function cloudGetReposForProjects(client: BitbucketClient, projects: strin
 
         logger.debug(`Fetching all repos for project ${project} for workspace ${workspace}...`);
         try {
-            const path = `/repositories/${workspace}` as CloudGetRequestPath;
-            const repos = await getPaginatedCloud<CloudRepository>(path, async (url) => {
-                const response = await client.apiClient.GET(url, {
+            const repos = await getPaginatedCloud<CloudRepository>(`/repositories/${workspace}` as CloudGetRequestPath, async (url) => {
+                const { path, query } = parseUrl(url, client.baseUrl);
+                const response = await client.apiClient.GET(path, {
                     params: {
+                        path: {
+                            workspace,
+                        },
                         query: {
+                            ...query,
                             q: `project.key="${project_name}"`
                         }
                     }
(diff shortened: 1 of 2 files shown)


WendelBartzUbots:
@bardock I tested and found the following error:

[dev:backend ] 2025-09-10T18:08:08.875Z error: [bitbucket] Failed to get repos for workspace facebot: TypeError: Invalid URL
[dev:backend ] 2025-09-10T18:08:08.876Z error: [connection-manager] Failed to compile repo data for connection 1 (starter-connection): TypeError: Invalid URL

Apparently, the baseUrl is not being used, and when adjusting it, the issue of duplicating /2.0 occurs.

packages/backend/src/bitbucket.ts
```diff
@@ -170,21 +170,33 @@ const getPaginatedCloud = async <T>(
     }
     return results;
 }
-   
+
+/**
+ * Parse the url into a path and query parameters to be used with the api client (openapi-fetch)
+ */
+function parseUrl(url: string, baseUrl: string): { path: string; query: Record<string, string>; } {
```
brendan-kellam: Since we are using this in all `get` callbacks passed to `getPaginatedCloud` to handle `response.next`, I wonder if we can just move this logic into `getPaginatedCloud` directly?

For example, the signature could be:
```TypeScript
const getPaginatedCloud = async <T>(
    path: CloudGetRequestPath,
    get: (path: CloudGetRequestPath, query?: string) => Promise<CloudPaginatedResponse<T>>
): Promise<T[]>
```

And `getPaginatedCloud` handles deconstructing `response.next` into `path` and optionally `query`.

packages/backend/src/bitbucket.ts
```diff
@@ -170,21 +170,33 @@ const getPaginatedCloud = async <T>(
...
+    const fullUrl = new URL(url);
```
brendan-kellam, line 181: > @bardock I tested and found the following error:
> 
> [dev:backend ] 2025-09-10T18:08:08.875Z error: [bitbucket] Failed to get repos for workspace facebot: TypeError: Invalid URL [dev:backend ] 2025-09-10T18:08:08.876Z error: [connection-manager] Failed to compile repo data for connection 1 (starter-connection): TypeError: Invalid URL
> 
> Apparently, the baseUrl is not being used, and when adjusting it, the issue of duplicating /2.0 occurs.

RE @WendelBartzUbots @bardock - best guess is that `url` is `/repositories/${workspace}` (and not the full url) for the first call of the callback provided to `getPaginatedCloud`. I think the approach I outlined above should fix this (i.e., just deconstruct `response.next` and not `path`)

packages/backend/src/bitbucket.ts
bardock: Sorry guys, there was a bug in my code due to a change I had made at the last moment and I was sure I had tested it.

I made the modifications that @brendan-kellam suggested. I think it looks cleaner, the only issue is that the path was left with `/2.0` at the beginning and I had no other option but to remove it in a hacky way with a regex.
Another option is to always keep the original path fixed and only get the query params from the next url, but that's not ideal either.

@WendelBartzUbots can you try again please?

Let me know if you see anything else to adjust or if you prefer another approach. 
Thanks!

brendan-kellam: Kk this LGTM thanks 👍 Once @WendelBartzUbots confirms, this lgtm to merge. Could you also add a changelog entry? thx!

bardock: Sure, thx! Added the changelog entry under the "Unreleased" section.

WendelBartzUbots: I tried testing again here but couldn’t manage, I need to spend more time on it and unfortunately I don’t have any :( Especially because even if I fix this bug, I realized I have another issue when trying to pull the repositories.

brendan-kellam: No worries - let's merge this in for now and we can reopen if other issues arise.

@WendelBartzUbots please feel free to raise a new issue for the problem with pulling repositories.

brendan-kellam (approved):

PR #508, opened 2025-09-17
[experimental] feat(ee): GitHub permission syncing

brendan-kellam:
This PR adds GitHub permission syncing behind a experimental flag. From the docs:

> Permission syncing allows you to sync Access Permission Lists (ACLs) from a code host to Sourcebot. When configured, users signed into Sourcebot (via the code host’s OAuth provider) will only be able to access repositories that they have access to on the code host. Practically, this means:
> 
> - Code Search results will only include repositories that the user has access to.
> - Code navigation results will only include repositories that the user has access to.
> - Ask Sourcebot (and the underlying LLM) will only have access to repositories that the user has access to.
> - File browsing is scoped to the repositories that the user has access to.

It can be enabled by setting `EXPERIMENT_EE_PERMISSION_SYNC_ENABLED=true`. As this is a EE feature, a [license key](https://docs.sourcebot.dev/docs/license-key#license-key) is required.

## How it works 🧠 dump
- In the database, we maintain a `UserToRepoPermission` join table between `User` and `Repo`. This can be thought of as an edge in a graph.
- If there is an edge between a given User and Repo, then it indicates that the user has access to that given repo.
- At query time, in `prisma.ts` we use a [Prisma client extension](https://www.prisma.io/blog/client-extensions-preview-8t3w27xkrxxn) to extend any queries that operate on the `Repo` table and add an additional `where` statement that enforces a edge must exist between the repo and _current_ user.  This statement looks like:
```TypeScript
OR: [
    // Only include repos that are permitted to the user,
    {
        permittedUsers: {
            some: {
                userId,
            }
        }
    },
    // or are public.
    {
        isPublic: true,
    }
]
```
- When I refer to "current user", I mean the user that is making the authenticated request for the given action (authentication context). E.g., if the signed in user is bob, and the client calls the `getRepos()` action, then the current user is bob.
- This extended Prisma client is exposed on `withAuthV2` and `withOptionalAuthV2`. My mental model here is that the prisma client exposed via this middleware is scoped to the authentication context and can be used "safely" (i.e., we don't need to worry about always having a `where` clause on all `Repo` queries).
- Once we migrate over to v2 (next up), we should mark the global `prisma` object as `__unsafe` to explicitly denote that this user auth context scoping is **not** applied. There will still be parts of the code where this unscoped client will be needed (e.g., initialize code), but most everything else can use the scoped client (all actions, api routes, etc.)
- **<-- how is the `UserToRepoPermission` table hydrated? -->**
- On the worker side, this PR adds two new schedulers: `UserPermissionSyncer` and `RepoPermissionSyncer`. These workers periodically reach out to the code host to determine the repos <--> user mapping. These schedulers both build this mapping in parallel (and overlap), but in different directions:
   - **Repo driven**: `RepoPermissionSyncer` fetches a list of all users that can view a given repository. For GitHub, it uses the [
/repos/{owner}/{repo}/collaborators](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators) api. The authentication token in the `config.json` is used to do this.
   - **User driven**: `UserPermissionSyncer` fetches a list of all repositories that a given user has access to. For GitHub, uses [/user/repos](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-the-authenticated-user). The **access token** associated with the User's GitHub account (stored in the `Account` table and obtained when the user signed in with the OAuth GitHub app) is used to authenticate.
- These workers process these as asynchronous jobs that are queued using Redis and BullMQ (similar to repository indexing and connection syncing).
- The default sync interval is 24 hours for both syncers (e.g., a given User will have a User Driven permission sync job run once every 24 hours).
- **<-- Why do we need two syncers? Wouldn't it be sufficient to just have a single direction? -->**
- Consider the following scenarios:
   1. **Only repo driven**: assume all repos have finished syncing now. If a user signs up afterwards, no `UserToRepoPermission` entries linking the user and the repos will exist, so the user will not see any repos until the next repo driven sync (24 hours).
   2. **Only user driven**: it's the inverse. Assume all users have finished syncing now. If a repo is added, no `UserToRepoPermission` entries linking the repo and the users will exist, so the repo will not be discovered until the next user driven sync (24 hours).
- There are probably other schemes out there to solve this issue, but this was the most straight forward at the time.

## Summary by CodeRabbit

- New Features
  - Experimental permission syncing (Enterprise): sync repository access from GitHub to Sourcebot. Enable via EXPERIMENT_EE_PERMISSION_SYNC_ENABLED; defaults to 24h user/repo sync intervals. Currently supports GitHub.
  - Improved repository visibility handling with explicit public/private flag.
  - Simplified app and API calls for browsing, files, images, and search—no domain headers/params required.

- Documentation
  - New “Permission syncing” guide, configuration and environment variable updates, navigation additions, reusable experimental warning, and Agents overview update (adds “Review Agent” card).

- Chores
  - Added dev script: dev:prisma:generate.


brendan-kellam:
@coderabbitai review

docs/snippets/experimental-feature-warning.mdx
```diff
@@ -0,0 +1,4 @@
+
+<Warning>
+This is an experimental feature. Certain functionality may be buggy or incomplete, and breaking changes may ship in non-major releases. Have feedback? Submit a [issue](https://github.com/sourcebot-dev/sourcebot/issues) on GitHub.
```
msukkari: nit: is there a more professional way of replacing "things may be buggy"? I think just saying it's experimental and certain functionality may be incomplete is sufficient

packages/backend/src/ee/repoPermissionSyncer.ts
```diff
@@ -0,0 +1,301 @@
...
+                                    }
+                                }
+                            }
+                        },
+                    ]
+                }
+            });
+
+            await this.schedulePermissionSync(repos);
+        }, 1000 * 5);
```
msukkari, line 101: nit: minimum of this number and `experiment_repoDrivenPermissionSyncIntervalMs` ?

packages/db/prisma/schema.prisma
```diff
@@ -51,12 +51,18 @@ model Repo {
   indexedAt          DateTime?
   isFork             Boolean
   isArchived         Boolean
+  isPublic           Boolean?
```
msukkari: why is this needed? if the repo is public wouldn't everyone have perms to see it?

packages/db/prisma/schema.prisma
```diff
@@ -51,12 +51,18 @@ model Repo {
...
   webUrl             String?
   connections        RepoToConnection[]
   imageUrl           String?
-  repoIndexingStatus RepoIndexingStatus @default(NEW)
+  repoIndexingStatus RepoIndexingStatus     @default(NEW)
+  permittedUsers     UserToRepoPermission[]
+
+  permissionSyncJobs RepoPermissionSyncJob[]
+  /// When the permissions were last synced successfully.
+  permissionSyncedAt DateTime?
```
msukkari: nit: not a big deal but couldn't this be fetched from the new `RepoPermissionSyncJob` table? risk of data inconsistencies

packages/db/prisma/schema.prisma
```diff
@@ -290,6 +314,41 @@ model User {
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
+
+  permissionSyncJobs UserPermissionSyncJob[]
+  permissionSyncedAt DateTime?
```
msukkari, line 313: same comment as in `Repo` table

packages/backend/src/ee/repoPermissionSyncer.ts
```diff
@@ -0,0 +1,301 @@
...
+                },
+            });
+            logger.error(errorMessage(repo.displayName ?? repo.name));
+        } else {
+            logger.error(errorMessage('unknown repo (id not found)'));
+        }
+    }
+}
+
+const getFirstConnectionWithToken = (repo: RepoWithConnections) => {
```
msukkari: this looks weird - are the different config casts needed / I thought we did this somewhere else too?

packages/backend/src/types.ts
```diff
@@ -50,4 +51,6 @@ export type DeepPartial<T> = T extends object ? {
 } : T;
 
 // @see: https://stackoverflow.com/a/69328045
-export type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
\ No newline at end of file
+export type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
+
+export type RepoWithConnections = Repo & { connections: (RepoToConnection & { connection: Connection })[] };
```
msukkari, line 56: nit: we also define this type in repoManager, can you change the import there?

packages/backend/src/ee/repoPermissionSyncer.ts
```diff
@@ -0,0 +1,301 @@
...
+                },
+                opts: {
+                    removeOnComplete: env.REDIS_REMOVE_ON_COMPLETE,
+                    removeOnFail: env.REDIS_REMOVE_ON_FAIL,
+                }
+            })))
+        });
+    }
+
+    private async runJob(job: Job<RepoPermissionSyncJob>) {
```
msukkari, line 133: discussed offline: we should also check visibility of repo here and update `isPublic` flag accordingly (or some other solution)

msukkari: conceptually we should make sure that the repo perms (including `isPublic` info) should be updated on this sync interval (as opposed to being only tied to connection sync)

msukkari (dismissed):

packages/backend/src/ee/repoPermissionSyncer.ts
brendan-kellam: tbh I can't imagine the repoDrivenPermissionSyncIntervalMs would be set to < 5s otherwise you would hit the 5000 GitHub rate limit at ~7 repositories, so going to leave as is for now.

(`60s / 5s = 12 times syncs per minute, 12 syncs * 60 minutes = 720 syncs per hour, 5000 api requests per hour / 720 syncs per hour = 6.9 repos`)

packages/backend/src/ee/repoPermissionSyncer.ts
brendan-kellam: good callout - going to combine this with the existing path in repoManager s.t., there is a single source of truth for getting credentials for a repository.

packages/db/prisma/schema.prisma
brendan-kellam: discussed offline

packages/db/prisma/schema.prisma
brendan-kellam: Yes it could, but going to leave the field since it makes queries simpler & more performant

packages/db/prisma/schema.prisma
brendan-kellam: see other comment

packages/backend/src/ee/repoPermissionSyncer.ts
brendan-kellam: heads up: this was slightly more complicated than we previously thought, so going to leave it as is more now and we can comeback to it if it comes up

PR #515, opened 2025-09-17
chore(web): Remove deprecated connection creation/edit UI

msukkari:
This is a remnant of our previous SaaS offering and should be deleted to remove confusion

New UI:

image
image


brendan-kellam:
cc @weberpatr we decided to deprecate this feature

PR #532, opened 2025-09-21
fix(web): Change buttons into Links in various places

### LINKED ISSUE #519: [FR] Cmd/Ctrl clicking a file should open it in a new tab in the code browser
Usecase:
a user may perform a search and then want to browse the full code for a number of the search result in separate tabs. Currently this is not easy to do.

I would be good to allow for opening a file (with the view centered on the relevant search chunk) in a new tab.
This could be added when the user cmd/ctrl + clicks on the file name? That or there could be a separate button with the "open in new tab" icon alongside the "preview" button:
https://thenounproject.com/browse/icons/term/new-window/

brendan-kellam:
Converted the following things into links:
- File headers (including breadcrumb dropdown)
- Files & directories in the file tree
- Reference list

Fixes #519 

The last remaining area I would like to turn into links is the code nav buttons (find all refs / defs). These are slightly more complicated, so I'll leave them for a separate PR.


KonradStanski:
Thanks!!!!

PR #533, opened 2025-09-22
fix(azure): add azure devops http header (git clone & git fetch)

### LINKED ISSUE #522: [BUG] Azure DevOps repository cloning fails due to incorrect authentication method  Description
### Describe the bug

When connecting to an Azure DevOps instance, the initial API connection and repository discovery succeed, but the subsequent git clone operation fails with an Authentication failed error.

This issue stems from the way repoManager constructs the clone URL for Azure DevOps repositories.

**Logs**

```
The initial sync job shows a successful API connection:

[backend] | 2025-09-19T06:42:54.369Z info: [connection-manager] Connection sync job for connection azure_devops (id: 5, jobId: 17) completed
```

However, the cloning process initiated by repoManager fails:

```
[backend] | 2025-09-19T06:42:54.796Z error: [repo-manager] Failed to sync repository [internal-devops-host]/mpilot_perception/auto-tagging-platform ... Error: Failed to clone repository ...
[backend] | fatal: could not read Password for 'https://ewfdprbvpjgbntuinx76blr2ikdllcxuxb26chlumgffjofrltsq@[internal-devops-host]': No such device or address
```

**Root Cause Analysis**

Incorrect Credential Formatting: In packages/backend/src/repoManager.ts, the getCloneCredentialsForRepo function returns only a password (the PAT) for azuredevops connections.
Incorrect URL Construction: The syncGitRepository function in the same file has a fallback mechanism that incorrectly assigns the password to the username field of the clone URL when username is not provided. This results in an improperly formatted URL like https://<PAT>@[internal-devops-host]/....
Azure DevOps Authentication Mechanism: Azure DevOps does not support authentication via a PAT embedded in the username part of a URL. It requires the PAT to be supplied as the password in a Basic Authentication scheme, typically with an arbitrary (or empty) username.
Manual Verification

We have confirmed the correct authentication method via the command line.

Failed Attempt (mimicking current behavior):

```
git clone https://pat:YOUR_PAT@[internal-devops-host]/...
# Result: fatal: Authentication failed
```

Successful Attempt (correct method):
This method injects the credentials directly into the HTTP header, which is the correct way to authenticate with Azure DevOps.

```
TOKEN="YOUR_PAT"
AUTH_HEADER="Authorization: Basic $(echo -n ":$TOKEN" | base64)"
git -c http.extraHeader="$AUTH_HEADER" clone https://[internal-devops-host]/...
# Result: Success, cloning begins.
```

Suggested Solution

The repoManager needs to be updated to handle Azure DevOps authentication correctly.

Modify git.ts: The cloneRepository and fetchRepository functions in packages/backend/src/git.ts should be updated to accept an optional config parameter (a string array) to pass custom configurations like http.extraHeader to the underlying simple-git call.

Modify repoManager.ts: In syncGitRepository, add logic to detect if the repository is from an azuredevops connection.

If it is, construct the Authorization: Basic ... header using the PAT and pass it to cloneRepository/fetchRepository via the new config parameter.
If it is not, retain the existing logic of embedding credentials in the URL.
This will ensure that Azure DevOps repositories are cloned using the correct, header-based authentication method without affecting other Git providers.

### To reproduce

1. Configure a new code host connection of type azuredevops.
2. Provide the URL for your Azure DevOps server (e.g., https://[internal-devops-host]) and a valid Personal Access Token (PAT) with at least Code (Read) permissions.
3. Configure the connection to sync a specific repository.
4. Trigger a sync for the connection.
5. Observe the backend logs for the repo-manager service.

Sourcebot version (e.g. v3.0.1): 4.7.0

### Additional information

_No response_

Shaoqi-Cen:
Close #522 
This PR updates git clone and git fetch commands to support Azure DevOps repositories requiring http.extraHeader authentication. Testing confirms successful cloning and fetching of Azure DevOps repositories using this header-based authentication method.


msukkari:
Thanks @Shaoqi-Cen will take a look. Out of curiosity could you help me understand which scenarios this is required vs. the current PAT system? In my testing I was able to clone and fetch repos from ADO cloud with the current implementation

Shaoqi-Cen:
> Thanks @Shaoqi-Cen will take a look. Out of curiosity could you help me understand which scenarios this is required vs. the current PAT system? In my testing I was able to clone and fetch repos from ADO cloud with the current implementation

Thanks for confirming it works on ADO Cloud. After digging into our environment, we found the issue is environment-specific: our target Azure DevOps (behind IIS/with Windows auth enabled) advertises multiple auth schemes (WWW-Authenticate: Bearer, Basic, Negotiate, NTLM). When the PAT is embedded in the URL, Git/libcurl prefers NTLM over Basic; since PATs only work with Basic, the NTLM handshake fails and auth is rejected. Forcing Basic (by sending an explicit Authorization: Basic header) resolves the issue on our side.

So this change is required in environments where NTLM/Negotiate are enabled and prioritized (common on-prem or behind certain reverse proxies). In ADO Cloud, where Basic is effectively used, the current PAT flow continues to work. Thanks again for taking a look!

msukkari:
Sorry just to be clear, I tested the released version on ADO cloud and it worked. I haven't tested the implementation in this PR on ADO cloud yet.

Thanks for digging into the underlying reason why this is necessary in your case @Shaoqi-Cen . This is definitely something we should get merged in. However, I think the general approach should be a bit different to align better with how we like to handle url's across the different platforms. My recommendation:

- modify `getCloneCredentialsForRepo` to read the `deploymentType` in the `AzureDevOpsConnectionConfig` object. If the deploymentType is `server` then return null so the PAT doesn't get embedded in the url
- add a new param to `fetchRepository` and `cloneRepository` for the auth header. If this is null then go through the existing path. If the authHeader is non null, pass that into the call

Should be the same functionality but slightly cleaner to prevent the git helpers from needing to know the underlying code host platforms. If my explanation isn't clear let me know - happy to make the change myself

PR #543, opened 2025-09-27
fix(ado): Manually pass token through http header for ado server

msukkari:
Thank you @Shaoqi-Cen for proposing the solution for this and providing extremely useful context in https://github.com/sourcebot-dev/sourcebot/pull/533

This PR implements the same underlying logic but with the proposed structure. Putting up this PR so we can get this fix merged asap

This PR also makes the `deploymentType` field in the ADO config mandatory. This makes it much easier to tell with auth strategy to use. This will technically break deployments which don't pass this field in, but the error logs are very clear here and the behavior of the deployment won't change after the deploymentType is set, so I don't think it's worth pushing a major version 

## Summary by CodeRabbit

- **New Features**
  - Optionally send Azure DevOps Server auth token via HTTP header for clone/fetch operations.
- **Bug Fixes**
  - Fixed authentication flow for Azure DevOps Server deployments.
- **Documentation**
  - Updated Azure DevOps connection examples to include deploymentType ("cloud" or "server").
  - Changelog entry added noting the auth token handling fix.
- **Chores**
  - Schemas tightened: deploymentType is now required (no default) for Azure DevOps connections.

### DIFF
--- a/packages/backend/src/utils.ts
+++ b/packages/backend/src/utils.ts
@@ -193,19 +193,31 @@ export const getAuthCredentialsForRepo = async (repo: RepoWithConnections, db: P
             const config = connection.config as unknown as AzureDevOpsConnectionConfig;
             if (config.token) {
                 const token = await getTokenFromConfig(config.token, connection.orgId, db, logger);
-                return {
-                    hostUrl: config.url,
-                    token,
-                    cloneUrlWithToken: createGitCloneUrlWithToken(
-                        repo.cloneUrl,
-                        {
-                            // @note: If we don't provide a username, the password will be set as the username. This seems to work
-                            // for ADO cloud but not for ADO server. To fix this, we set a placeholder username to ensure the password
-                            // is set correctly
-                            username: 'user',
-                            password: token
-                        }
-                    ),
+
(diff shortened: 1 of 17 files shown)


msukkari:
@coderabbitai review

packages/backend/src/utils.ts
```diff
@@ -193,19 +193,31 @@ export const getAuthCredentialsForRepo = async (repo: RepoWithConnections, db: P
...
-                    ),
+
+                // For ADO server, multiple auth schemes may be supported. If the ADO deployment supports NTLM, the git clone will default
+                // to this over basic auth. As a result, we cannot embed the token in the clone URL and must force basic auth by passing in the token
+                // appropriately in the header. To do this, we set the authToken field here
+                if (config.deploymentType === 'server') {
+                    return {
+                        hostUrl: config.url,
+                        token,
+                        authToken: "Authorization: Basic " + Buffer.from(`:${token}`).toString('base64')
```
brendan-kellam: authHeader

brendan-kellam (approved):

msukkari:
@coderabbitai review

PR #577, opened 2025-10-28
feat(security): Add env var to restrict api key creation

msukkari:
Adds experimental env var to prevent non-admins to create API Keys: `EXPERIMENT_DISABLE_API_KEY_CREATION_FOR_NON_ADMIN_USERS`. Allows teams to control usage of Sourcebot with MCP server.

In the future we'll want to have a more useful dashboard to be able to audit API keys, scope, and usage.


brendan-kellam (approved):
LGTM

PR #563, opened 2025-10-15
chore(worker,web): Repo indexing stability improvements + perf improvements to web

brendan-kellam:
image

_...well this PR escalated quickly 👀_

TL;DR: solved some repo indexing concurrency issues by moving from BullMQ to [GroupMQ](https://openpanel-dev.github.io/groupmq/), refactored how we represent indexing jobs in the db, and generally deleted a ton of things that got in the way of web perf.

## Where it started

We were getting bug reports (#523, #436) that repo indexing would fail at the clone or fetch stage with strange git errors. After some investigation on a test deployment of ~1k repos, we noticed that the root cause seemed to be that multiple indexing jobs could run on the **same repository** at the same time. When this happens, two workers will attempt to perform git operations on the same folder, resulting in the weird git behaviour we were seeing. For example:

1. Worker 1 scheduled on `repoid=1`
1. Worker 2 scheduled on `repoid=1`
1. Worker 1 starts. Spawns git process to clone into `/repos/1`
1. Worker 1 pre-empted by Node.JS runtime.
1. Worker 2 starts. Spawns git process to clone into `/repos/1`
<-- now two git processes are operating on the same repo. bad news -->
1. Worker 2 receives error from git process that clone failed. Exists with job failed.
1. Worker.1 pre-empted by Node.JS runtime.
1. Worker 1 receives error from git process that clone failed. Exists with job failed.

**Q**: Why were two workers being scheduled for the same repository in the first place?
**A**: I'm still not certain, but here's my theory: The job scheduler uses the the database in order to schedule new indexing jobs. This happens in two operations: 1) fetch all repositories that need indexing, and 2) create BullMQ jobs for these repositories and flag them as indexing. We **were not** performing 1 and 2 in a transaction, so this was **not** an atomic operation. I noticed that failures would typically happen in our k8s cluster when a new replica was being spun up. Likely, we were getting interleavings between replica A and B something like `A:1, B:1, A:2, B:2`, resulting in duplicate jobs for the same set of repos.

Put another way, we were using the Postgres database as a distributed lock to enforce the "1 repo per worker" invariant. This felt error prone - what we really want is the invariant to be enforced at the **queue** level within Redis.

BullMQ has the [concept of groups](https://docs.bullmq.io/bullmq-pro/groups):

> Groups allows you to use a single queue while distributing the jobs among groups so that the jobs are processed one by one relative to the group they belong to.

[...]

PR #579, opened 2025-10-29
feat(worker,web): Improved connection management

brendan-kellam:
Similar to #563 and #572, this PR does two things:
1. Adds a separate `ConnectionSyncJob` table for job tracking. Also moves to using groupmq.
2. Improves the connections table with information about connection sync jobs.

### Screenshots
_Connection list view_
image

_Connection details view_
image

_Hovering over a warning icon_
image

_Hovering over a error icon_
image

_Info banner in repository page when no repos are configured_
image

_Notification dot when there is a connection that is syncing for the same time_
image

_Notification dot in details view_
image

_Notification dot in repos table to indicate first time indexing_
image

TODO:
- [ ] We can probably fully remove BackendError

## Release Notes

* **New Features**
  * Added connection management interface in settings with detailed sync job history and status tracking.
  * Introduced first-time sync notification indicators for repositories and connections.
  * Enhanced error reporting with detailed sync warnings and diagnostics.

* **Improvements**
  * Strengthened GitHub App configuration schema with improved validation for deployment settings and private key handling.
  * Added connection synchronization statistics and monitoring capabilities.

### DIFF
--- a/packages/web/src/app/[domain]/settings/layout.tsx
+++ b/packages/web/src/app/[domain]/settings/layout.tsx
@@ -102,7 +102,7 @@ export default async function SettingsLayout(
             {
                 title: "Connections",
                 href: `/${domain}/settings/connections`,
-                hrefRegex: `/${domain}/settings/connections(\/[^/]+)?$`,
+                hrefRegex: `/${domain}/settings/connections(/[^/]+)?$`,
                 isNotificationDotVisible: connectionStats.numberOfConnectionsWithFirstTimeSyncJobsInProgress > 0,
             }
         ] : []),


brendan-kellam:
@coderabbitai review

packages/backend/src/env.ts
```diff
@@ -56,6 +56,8 @@ export const env = createEnv({
 
         EXPERIMENT_EE_PERMISSION_SYNC_ENABLED: booleanSchema.default('false'),
         AUTH_EE_GITHUB_BASE_URL: z.string().optional(),
+
+        FORCE_ENABLE_ANONYMOUS_ACCESS: booleanSchema.default('false'),
```
msukkari: why is this being added here?

packages/backend/src/repoIndexManager.ts
```diff
@@ -149,7 +149,8 @@ export class RepoIndexManager {
     }
 
     private async scheduleCleanupJobs() {
-        const thresholdDate = new Date(Date.now() - this.settings.repoGarbageCollectionGracePeriodMs);
+        const gcGracePeriodMs = new Date(Date.now() - this.settings.repoGarbageCollectionGracePeriodMs);
```
msukkari, line 152: out of curiosity why is this needed? is there some sort of race condition happening between indexing and cleanup?

packages/backend/src/env.ts
brendan-kellam: Good callout not needed

packages/backend/src/repoIndexManager.ts
brendan-kellam: I actually don't recall why we added this, but according to the schema: "The grace period (in milliseconds) for garbage collection. Used to prevent deleting shards while they're being loaded. Defaults to 10 seconds."

PR #585, opened 2025-10-30
feat(ee): GitLab permission syncing

brendan-kellam:
This PR adds support for GitLab permission syncing.

## Summary by CodeRabbit

* **New Features**
  * Added GitLab support for repository and user permission synchronization
  * Introduced GitLab environment configuration settings
  * Enhanced permission sync job logging for improved traceability
  * Conditional OAuth scopes for GitLab SSO when permission syncing is enabled

### DIFF
--- a/packages/backend/src/gitlab.ts
+++ b/packages/backend/src/gitlab.ts
@@ -305,7 +305,8 @@ export const getProjectsForAuthenticatedUser = async (visibility: 'private' | 'i
             } : {}),
             perPage: 100,
         });
-        return fetchWithRetry(fetchFn, `authenticated user`, logger) as Promise<ProjectSchema[]>;
+        const response = await fetchWithRetry(fetchFn, `authenticated user`, logger);
+        return response;
     } catch (error) {
         Sentry.captureException(error);
         logger.error(`Failed to fetch projects for authenticated user.`, error);


brendan-kellam:
@coderabbitai review

packages/backend/src/ee/userPermissionSyncer.ts
```diff
@@ -183,6 +189,37 @@ export class UserPermissionSyncer {
...
+                    if (!account.access_token) {
+                        throw new Error(`User '${user.email}' does not have a GitLab OAuth access token associated with their GitLab account.`);
+                    }
+
+                    const api = await createGitLabFromOAuthToken({
+                        oauthToken: account.access_token,
+                        url: env.AUTH_EE_GITLAB_BASE_URL,
+                    });
+
+                    // @note: we only care about the private and internal repos since we don't need to build a mapping
```
msukkari, line 203: might be a noob question but gpt says that internal repos are repos that are accessible to anyone who's signed into the gitlab instance. If that's the case, aren't those repos technically "public" (and thus no mapping is needed for them)?

msukkari (dismissed):

packages/backend/src/ee/userPermissionSyncer.ts
brendan-kellam: I think it's best to have the same mapping for what we view as public in our DB be the same as what the code host views as public.

PR #588, opened 2025-10-31
fix(web): Fix `/settings/connections` throwing a error when there is a `git` connection present

### LINKED ISSUE #586: [bug] Git connections defined using v3 config schema don't load in UI
### Describe the bug

Git connections defined in config.json cause errors and don't load in the UI. The connectionType field gets set to "git" instead of "generic-git-host", causing the application to fail when processing these connections.

[packages/backend/src/configManager.ts:85](https://github.com/sourcebot-dev/sourcebot/blob/4899c9fbc755851af2ddcce99f4a4200f2faa4f6/packages/backend/src/configManager.ts#L85) assigns newConnectionConfig.type (which is `"git"` from the v3 config schema) directly to the `connectionType` field, but the codebase expects `"generic-git-host"`.

### To reproduce

1. Add a git connection to config.json:

```
{
  "connections": {
    "my-repos": {
      "type": "git",
      "url": "file:///repos/my-org/*"
    }
  }
}
```

2. Attempt to load the connections view in a browser.

Sourcebot version (e.g. v3.0.1): v4.8.1

### Additional information

I have applied the following patch to my local build in the meantime to get around the issue:

```
diff --git a/packages/backend/src/configManager.ts b/packages/backend/src/configManager.ts
index 14c6b0d..2e5ca51 100644
--- a/packages/backend/src/configManager.ts
+++ b/packages/backend/src/configManager.ts
@@ -82,7 +82,7 @@ export class ConfigManager {
                         data: {
                             name: key,
                             config: newConnectionConfig as unknown as Prisma.InputJsonValue,
-                            connectionType: newConnectionConfig.type,
+                            connectionType: newConnectionConfig.type === 'git' ? 'generic-git-host' : newConnectionConfig.type,
                             isDeclarative: true,
                             org: {
                                 connect: {
```

brendan-kellam:
This PR fixes a exception that was being thrown when viewing the connections page with a `git` connection. The root-cause was that we were calling `getCodeHostIcon` with a `connectionType`, where it was meant to accept a `codeHostType`.

## Backstory

The `Connection` table has a `connectionType` column that stores the connection's type (i.e., the code host). This field is sourced from the config schema and can be one of the following values:
```
github
gitlab
gitea
gerrit
bitbucket
azuredevops
git
```

The `Repo` table has a `external_codeHostType` that stores the type of code host it belongs to. This is deceptively similar to `connectionType`, _but for reasons lost to time_, the values this field can take on are slightly different:
```
github
gitlab
gitea
gerrit
bitbucket-server
bitbucket-cloud
generic-git-host
azuredevops
```

Both `connectionType` and `external_codeHostType` were of type `String`, so we didn't have any type guarantees, and there were some places where we were comparing to the wrong values (hence this bug).

To help combat this, I've introduced two enums into the schema `ConnectionType` and `CodeHostType` that explicitly define the valid values for `connectionType` and `external_codeHostType`, respectively. Since we know all possible values for these fields, the migrations I've added simply map from the string into the corresponding enum value. This adds type safety throughout a lot of paths dealing with these fields.

Fixes #586 

## Summary by CodeRabbit

* **Refactor**
  * Strengthened type safety for code hosting platforms and repository connections through improved internal type definitions.
  * Migrated repository and connection data to use standardized enumerated types in the database for enhanced consistency.
  * Streamlined component rendering logic for repository and code host information display across the platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [ask sb] Fixed issue where reasoning tokens would appear in `text` content for openai compatible models. [#582](https://github.com/sourcebot-dev/sourcebot/pull/582)
 - Fixed issue with GitHub app token tracking and refreshing. [#583](https://github.com/sourcebot-dev/sourcebot/pull/583)
 - Fixed "The account is already associated with another user" errors with GitLab oauth provider. [#584](https://github.com/sourcebot-dev/sourcebot/pull/584)
+- Fixed error when viewing a generic git connection in `/settings/connections`. [#588](https://github.com/sourcebot-dev/sourcebot/pull/588)
 
 ## [4.8.1] - 2025-10-29
 


brendan-kellam:
@coderabbitai review

msukkari (approved):

PR #592, opened 2025-10-31
chore(tech-debt): Remove built-in secret manager

msukkari:
The built-in secret manager is an artifact of our deprecated cloud offering. It hasn't been supported for a while, and is unneeded tech debt. This PR completely removes the internal secret manager.

Tested the migration path and it works without errors when the Secrets table is empty. Surprisingly, it also seems to work when a secret is present (it just nukes the Secret table).

If you came to this PR because you've tried updating your Sourcebot deployment and are hitting issues with the DB migration, you will need to manually delete the Secret table in your DB. If you're using the internal DB then a clean wipe of your Sourcebot cache and restart should allow the migration to happen automatically. You'll need to switch to passing in secrets through environment variables (see [docs](https://docs.sourcebot.dev/docs/configuration/config-file) for more details).


brendan-kellam:
image
love to see it

brendan-kellam (approved):
LGTM 👍

PR #595, opened 2025-11-03
feat(ee): Add ability to link external accounts

msukkari:
Adds a new `identityProvider` top level object to the config file. This allows you to define one or more identity providers which will be used by Sourcebot.

There are two different types of identity providers: `sso` and `integration`. `sso` providers are used to auth, while `integration` providers are linked after authentication and are (currently) used for permission syncing. This allows users to login with a different identity provider than their code host platform (ex. login using `Okta` but still be able to connect to `GitHub` for permission syncing). It also sets the foundation to be able to oauth into multiple code host platforms to support permission syncing across multiple code host platforms

TODO:

- [x] docs
- [x] add oauth token refresh logic (probably in a different PR)

image
image

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added identity provider management with support for multiple providers (GitHub, GitLab, Google, Okta, Keycloak, Microsoft Entra ID, GCP IAP)
  * Added permission syncing feature allowing users to link and manage integration accounts
  * New "Linked Accounts" settings page for managing connected provider accounts
  * Added OAuth token refresh capability for seamless account synchronization

* **Bug Fixes**
  * Updated GitHub App configuration identifier for consistency

* **Configuration**
  * Expanded root configuration schema to support identity providers collection
  * Configuration path parameter is now optional

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 - [Experimental][Sourcebot EE] Added GitLab permission syncing. [#585](https://github.com/sourcebot-dev/sourcebot/pull/585)
+- [Sourcebot EE] Added external identity provider config and support for multiple accounts. [#595](https://github.com/sourcebot-dev/sourcebot/pull/595)
 
 ### Fixed
 - [ask sb] Fixed issue where reasoning tokens would appear in `text` content for openai compatible models. [#582](https://github.com/sourcebot-dev/sourcebot/pull/582)


docs/snippets/schemas/v3/app.schema.mdx
```diff
@@ -8,7 +8,7 @@
       "type": "object",
       "properties": {
         "type": {
-          "const": "githubApp",
+          "const": "github",
```
msukkari, line 11: note that this is technically a breaking change but it was a mistake to name this `githubApp` IMO so we should fix this asap

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Removed
 - Removed built-in secret manager. [#592](https://github.com/sourcebot-dev/sourcebot/pull/592)
 
+## Removed
```
brendan-kellam: nit: repeated changelog

schemas/v3/identityProvider.json
```diff
@@ -0,0 +1,160 @@
...
+                "purpose": {
+                    "enum": ["sso", "integration"]
+                },
+                "clientId": {
+                    "$ref": "./shared.json#/definitions/Token"
+                },
+                "clientSecret": {
+                    "$ref": "./shared.json#/definitions/Token"
+                },
+                "baseUrl": {
```
brendan-kellam, line 21: Do we expect anything secret to be stored in the baseUrl? Ideally this could be a string so we are consistent with other parts of the config (namely connection and languageModel urls)

schemas/v3/identityProvider.json
```diff
@@ -0,0 +1,160 @@
...
+            },
+            "required": ["provider", "purpose", "clientId", "clientSecret"]
+        },
+        "GoogleIdentityProviderConfig": {
+            "type": "object",
+            "properties": {
+                "provider": {
+                    "const": "google"
+                },
+                "clientId": {
```
brendan-kellam, line 83: For auth providers without a `purpose`, is it assumed they are for `sso`?

packages/schemas/src/v3/authProvider.type.ts
```diff
@@ -0,0 +1,275 @@
+// THIS IS A AUTO-GENERATED FILE. DO NOT MODIFY MANUALLY!
+
+export type IdentityProviderConfig =
```
brendan-kellam: Could you use this type instead of redeclaring a separate `IdentityProvider` type?

packages/web/src/ee/features/sso/sso.ts
```diff
@@ -10,174 +9,260 @@ import { prisma } from "@/prisma";
...
+import { hasEntitlement, loadConfig } from "@sourcebot/shared";
+import { getTokenFromConfig } from "@sourcebot/crypto";
+import type { IdentityProvider } from "@/auth";
+import { GCPIAPIdentityProviderConfig, GitHubIdentityProviderConfig, GitLabIdentityProviderConfig, GoogleIdentityProviderConfig, KeycloakIdentityProviderConfig, MicrosoftEntraIDIdentityProviderConfig, OktaIdentityProviderConfig } from "@sourcebot/schemas/v3/index.type";
 
 const logger = createLogger('web-sso');
 
-export const getSSOProviders = (): Provider[] => {
-    const providers: Provider[] = [];
+export const getEEIdentityProviders = async (): Promise<IdentityProvider[]> => {
```
brendan-kellam, line 24: See [this comment](https://github.com/sourcebot-dev/sourcebot/pull/595/files#r2488602079). Using the schema generated one would be useful since it is a [discriminated union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) - you won't need to have the `as GitHubIdentityProviderConfig` type casts

packages/web/src/ee/features/sso/sso.ts
```diff
@@ -10,174 +9,260 @@ import { prisma } from "@/prisma";
...
+            providers.push({ provider: createMicrosoftEntraIDProvider(clientId, clientSecret, issuer), purpose: "sso" });
+        }
+        if (identityProvider.provider === "gcp-iap") {
+            const providerConfig = identityProvider as GCPIAPIdentityProviderConfig;
+            const audience = await getTokenFromConfig(providerConfig.audience);
+            providers.push({ provider: createGCPIAPProvider(audience), purpose: "sso" });
+        }
+    }
 
+    // @deprecate
```
brendan-kellam: nit: could we add a comment on why this is here (i.e., back compat).

packages/web/src/ee/features/sso/sso.ts
```diff
@@ -10,174 +9,260 @@ import { prisma } from "@/prisma";
...
-                        // @see: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-the-authenticated-user
-                        ...(env.EXPERIMENT_EE_PERMISSION_SYNC_ENABLED === 'true' && hasEntitlement('permission-syncing') ?
-                            ['repo'] :
-                            []
-                        ),
-                    ].join(' '),
-                },
-            },
-        }));
+        providers.push({ provider: createGitHubProvider(env.AUTH_EE_GITHUB_CLIENT_ID, env.AUTH_EE_GITHUB_CLIENT_SECRET, env.AUTH_EE_GITHUB_BASE_URL), purpose: "sso" });
```
brendan-kellam: nit: one edgecase bug is if the user were to configure their `clientId` and `clientSecret` (in the config) to point at `AUTH_EE_GITHUB_CLIENT_ID` and `AUTH_EE_GITHUB_CLIENT_SECRET` environment variables, then there would be two providers. Maybe we should only do the legacy path if `config.identityProviders` is undefined?

packages/web/src/ee/features/sso/sso.ts
brendan-kellam: nvm disregard I see why you have a seperate type. I don't think you need the type casts though regardless

brendan-kellam:
@coderabbitai review

packages/web/src/app/[domain]/layout.tsx
```diff
@@ -123,6 +125,42 @@ export default async function Layout(props: LayoutProps) {
         )
     }
 
+    if (hasEntitlement("permission-syncing")) {
+        const integrationProviderStates = await getIntegrationProviderStates();
+        if (isServiceError(integrationProviderStates)) {
```
Comment (brendan-kellam) [inline packages/web/src/app/[domain]/layout.tsx]: can probably just `throw new ServiceErrorException`

schemas/v3/identityProvider.json
brendan-kellam, line 21: Also, could the `baseUrl` for GitHub and GitLab be optional and default to `https://github.com` and `https://gitlab.com` respectively?

msukkari: Made it optional. Question question about "default" - are these actually enforced by the schema? I tried to mess around with avj but doesn't seem like the `default` fields in the schema are actually doing anything more than a suggestion

brendan-kellam: Discussed offline - `default` fields in the schema are just suggestions

packages/web/src/ee/features/sso/sso.ts
```diff
@@ -10,174 +9,260 @@ import { prisma } from "@/prisma";
...
+
+    return providers;
+}
+
+const createGitHubProvider = (clientId: string, clientSecret: string, baseUrl?: string): Provider => {
+    return GitHub({
+        clientId: clientId,
+        clientSecret: clientSecret,
+        enterprise: {
+            baseUrl: baseUrl,
```
brendan-kellam: Discussed offline: we should make sure we properly support the case where baseUrl is provided as https://github.com

packages/web/src/ee/features/permissionSyncing/tokenRefresh.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        const data = await response.json();
+
+        const result = {
+            accessToken: data.access_token,
+            refreshToken: data.refresh_token ?? null,
+            expiresAt: data.expires_in ? Math.floor(Date.now() / 1000) + data.expires_in : 0,
+        };
+
+        const { prisma } = await import('@/prisma');
```
brendan-kellam: nit: any reason for a inline import?

packages/web/src/ee/features/permissionSyncing/tokenRefresh.ts
```diff
@@ -0,0 +1,158 @@
...
+export async function refreshOAuthToken(
+    provider: string,
+    refreshToken: string,
+    userId: string
+): Promise<{ accessToken: string; refreshToken: string | null; expiresAt: number } | null> {
+    try {
+        const config = await loadConfig(env.CONFIG_PATH);
+        const identityProviders = config?.identityProviders ?? [];
+
+        const providerConfig = identityProviders.find(idp => idp.provider === provider);
```
brendan-kellam: In the situation where there are multiple identity providers of the same type, I think we will run into issues here right? We always get whatever identity provider appears first in the config?


Aside: does auth.js even support the situation where multiple accounts from the same auth provider are linked to the same User?

packages/web/src/auth.ts
```diff
@@ -176,13 +188,34 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
...
+        async jwt({ token, user: _user, account }) {
             const user = _user as User | undefined;
             // @note: `user` will be available on signUp or signIn triggers.
             // Cache the userId in the JWT for later use.
             if (user) {
                 token.userId = user.id;
             }
+
+            // When a user links a new account, store the tokens if it's an integration provider
+            if (account && hasEntitlement('permission-syncing')) {
```
brendan-kellam: It doesn't really matter right now, but wouldn't we want integrations to live under the `sso` entitlement (or it's own entitlement)?

packages/web/src/ee/features/permissionSyncing/tokenRefresh.ts
brendan-kellam: Discussed offline: we should uniquely identify providers based on the combination of the providers type and baseUrl. This is kind of similar to what we do for language models (`getLanguageModelKey` in `features/chat/utils.ts`)

brendan-kellam:
**EDIT: #600 resolves this by moving the join table to be between `Account` and `Repo`

Discussed offline: there is a bug atm where if permission syncing is enabled and one or more integrations are configured, then _private_ repositories that require signin via the integrations will **not** be accessible (until the next user driven permission sync job runs). This is because `userPermissionSyncer.ts` schedules new sync jobs when a new `User` object is created (and therefore will sync the primary `sso` account), but not when subsequent `Account` objects are added.

The hack fix is to reset the `permissionSyncedAt` field to `null` when a new account is linked. The scheduler will take care of scheduling a new user permission sync job. **Caveat:** This will only happen if the first job _did not fail_ due to how the scheduler logic works.

The longer term fix is to support on-demand job queuing s.t., the webapp could schedule a user permission sync job on account linking.

schemas/v3/identityProvider.json
msukkari: yeah only providers with purpose can be considered for non-sso purposes

packages/web/src/ee/features/sso/sso.ts
msukkari: made it so this path is only done when no IdP's are set in the config

packages/web/src/app/[domain]/layout.tsx
Comment (msukkari) [reply packages/web/src/app/[domain]/layout.tsx]: what's displayed in the UI if we just throw in the `layout.tsx`?

packages/web/src/ee/features/permissionSyncing/tokenRefresh.ts
msukkari: cursor loves adding this, removed

packages/web/src/ee/features/permissionSyncing/tokenRefresh.ts
msukkari: good catch, switched to mapping this on `accountProviderId` instead and looping through all of the configs for that provider when refreshing

packages/web/src/auth.ts
msukkari: good question, this is a bit of a entitlements matrix question which we currently don't have standardized. My assumption here is that if they have the `permssion-syncing` entitlement then they also have the `sso` entitlement

PR #604, opened 2025-11-06
fix: return truncated content when token limit exceeded in MCP search_code

waynesun09:
## Summary

Fixes token truncation behavior in the MCP `search_code` tool to return partial content instead of discarding files completely when token limits are exceeded.

## Problem

Currently, when a search result file would exceed the `maxTokens` limit, the entire file is discarded with only a truncation message returned. This provides no useful data to the user, especially problematic when the first file is too large.

**Example failure scenario:**
- User requests search with `maxTokens=6000`
- First result file contains 10K tokens
- Current behavior: Returns ONLY truncation message, no file data
- Result: User gets no information at all

## Solution

Modified the truncation logic in `packages/mcp/src/index.ts` to:

1. Calculate remaining token budget before breaking the loop
2. If meaningful space remains (>100 tokens), truncate the file content to fit
3. Append a clear truncation marker: `...[content truncated due to token limit]`
4. Add the truncated content to results
5. Continue to add the overall truncation message at the end

## Changes

**File**: `packages/mcp/src/index.ts` (lines 125-142)

**Before:**
```typescript
if ((totalTokens + tokens) > maxTokens) {
    isResponseTruncated = true;
    break;  // Discards the file completely
}
```

**After:**
```typescript
if ((totalTokens + tokens) > maxTokens) {
    const remainingTokens = maxTokens - totalTokens;
    
    if (remainingTokens > 100) {
        const maxLength = Math.floor(remainingTokens * 4);
        const truncatedText = text.substring(0, maxLength) + 
            "\n\n...[content truncated due to token limit]";
        
        content.push({
            type: "text",
            text: truncatedText,
        });
    }
    
    isResponseTruncated = true;
    break;
}
```

## Benefits

✅ Users receive partial data instead of nothing  
✅ Better debugging and analysis experience  
✅ More useful for AI-powered code analysis workflows  
✅ Consistent with expected truncation behavior  
✅ Maintains backward compatibility (still includes truncation message)

## Example Impact

**Scenario**: Search returns 50 files, each ~2K tokens, with `maxTokens=10000`

**Before**: Returns first 5 complete files (10K tokens), discards file #6 completely  
**After**: Returns first 5 complete files + truncated version of file #6

## Testing

- ✅ Verified token calculation logic (chars/4 approximation)
- ✅ Tested with various token limits (100, 1000, 10000, 150000)
- ✅ Confirmed truncation marker appears correctly
- ✅ Validated backward compatibility (truncation message still appended)

## Related Issues

This fix specifically addresses issues observed in AI agent workflows where large codebases trigger token limits on early results, providing zero useful data to analysis tasks.

---

**Commit**: c5b8fda  
**Branch**: `fix/mcp-search-truncation`

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -123,6 +123,22 @@ server.tool(
             const tokens = text.length / 4;
 
             if ((totalTokens + tokens) > maxTokens) {
+                // Calculate remaining token budget
+                const remainingTokens = maxTokens - totalTokens;
+
+                if (remainingTokens > 100) {  // Only truncate if meaningful space left
+                    // Truncate text to fit remaining tokens (tokens ≈ chars/4)
+                    const maxLength = Math.floor(remainingTokens * 4);
+                    const truncatedText = text.substring(0, maxLength) + "\n\n...[content truncated due to token limit]";
+
+                    content.push({
+                        type: "text",
+                        text: truncatedText,
+                    });
+
+                    totalTokens += remainingTokens;
+                }
+
                 isResponseTruncated = true;
                 break;
             }
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed issue where search results exceeding token limits would be completely discarded instead of returning truncated content. [#604](https://github.com/sourcebot-dev/sourcebot/pull/604)
+
 ## [1.0.7] - 2025-10-28
 - Updated API client to match the latest Sourcebot release. [#555](https://github.com/sourcebot-dev/sourcebot/pull/555)
 


packages/mcp/src/index.ts
```diff
@@ -123,6 +123,20 @@ server.tool(
...
+
+                if (remainingTokens > 100) {  // Only truncate if meaningful space left
+                    // Truncate text to fit remaining tokens (tokens ≈ chars/4)
+                    const maxLength = Math.floor(remainingTokens * 4);
+                    const truncatedText = text.substring(0, maxLength) + "\n\n...[content truncated due to token limit]";
+
+                    content.push({
+                        type: "text",
+                        text: truncatedText,
+                    });
```
brendan-kellam, line 137: I believe we should still be updating the `totalTokens`, i.e., `totalTokens += remainingTokens`, otherwise the maxTokens would no longer be accurate to what is actually returned

brendan-kellam:
LGTM, this approach makes sense. Thanks for the contribution 👍 Left one comment, once resolved happy to merge

brendan-kellam (approved):

PR #607, opened 2025-11-09
fix(worker): Run setInterval as blocking

brendan-kellam:
setInterval is not blocking, so if a callback takes longer than the polling interval to execute, than we can run into situations where multiple setInterval callbacks run in parallel.

This change adds `setIntervalAsync` that handles this. See: https://mottaquikarim.github.io/dev/posts/setinterval-that-blocks-on-await/

## Bug Fixes
* Fixed race condition in job schedulers that could cause concurrent execution conflicts, resulting in data inconsistencies and scheduling failures. Significantly improved backend system stability and reliability of background job processing. Ensures more dependable scheduler operation across all components, particularly during simultaneous operations and peak-load situations.


brendan-kellam:
@coderabbitai review

PR #614, opened 2025-11-12
feat(mcp): Add pagination and filtering to list_repos tool

TJReinert:
## Summary
Add pagination and filtering capabilities to the `list_repos` MCP tool, enabling users to efficiently browse large repository lists.

### Changes
- **Filtering**: Added `query` parameter to filter repositories by name or displayName (case-insensitive)
- **Pagination**: Added `pageNumber` (1-indexed, default: 1) and `limit` (default: 50) parameters for result pagination
- **Schema**: Created `listReposRequestSchema` in schemas.ts for request validation with proper Zod descriptors
- **Response Info**: Added pagination metadata in responses to help users navigate results (e.g., "Use pageNumber 3 to see more")
- **Documentation**: Updated README with new parameters and usage examples
  
### Motivation
Previously, the `list_repos` tool returned all repositories at once, which is inefficient for organizations with large repository counts. Without pagination, responses can become too large for LMs to ingest, failing or consuming unnecessary context tokens. This change allows users to:
- Search for specific repositories by name to reduce result set
- Paginate through results efficiently to stay within LM context budgets
- Control response size with configurable limits
- Get more focused, actionable results from the tool

### Implementation & Improvements
These changes are implemented entirely on the MCP side since the API is fast. This approach is the least invasive version of this feature. If it is palatable to move filtering and pagination to the API level for better performance, I'm happy to implement that.

### Testing
The tool has been tested with various scenarios:
- No inputs, defaults applied (page 1, size 50)
- Page 1 with limit=5 (returns first 5 results)
- Page 2 with limit=5 (returns next 5 results with pagination info)
- Out-of-range page (returns helpful message)
- Query filtering (filters by repository name)
- Invalid input validation (rejects pageNumber < 1)

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -193,6 +193,9 @@ server.tool(
             );
         }
 
+        // Sort alphabetically for consistent pagination
+        filtered.sort((a, b) => a.repoName.localeCompare(b.repoName));
+
         // Apply pagination
         const startIndex = (pageNumber - 1) * limit;
         const endIndex = startIndex + limit;
--- a/packages/mcp/CHANGELOG.md
+++ b/packages/mcp/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added pagination and filtering to `list_repos` tool to handle large repository lists efficiently and prevent oversized responses that waste token context. [#614](https://github.com/sourcebot-dev/sourcebot/pull/614)
+
 ## [1.0.8] - 2025-11-10
 
 ### Fixed


packages/mcp/src/index.ts
```diff
@@ -177,13 +183,42 @@ server.tool(
...
+        if (query) {
+            const lowerQuery = query.toLowerCase();
+            filtered = response.filter(repo =>
+                repo.repoName.toLowerCase().includes(lowerQuery) ||
+                repo.repoDisplayName?.toLowerCase().includes(lowerQuery)
+            );
+        }
+
+        // Apply pagination
+        const startIndex = (pageNumber - 1) * limit;
```
brendan-kellam, line 200: I think we will need some ordering guarantees here - the current `/api/repos` endpoint calls [getRepos()](https://github.com/sourcebot-dev/sourcebot/blob/e20d514569520b4e1ec495f85d130a28339bbc91/packages/web/src/actions.ts#L471), which does not have a `orderBy` param.

Since we are doing the pagination logic here already, I think we can also do the ordering here as well by sorting `filtered` alphabetically.

brendan-kellam:
Could you also add a entry to packages/mcp/CHANGELOG.md?

brendan-kellam:
Released in v1.0.9 - thanks for the PR!

PR #615, opened 2025-11-13
fix(web): Search performance improvements

brendan-kellam:
After benchmarking our search api, I noticed that the zod `parseAsync` call we were doing in `searchApi.ts` was contributing to a significant portion of the total search time (average 64.6% in my benchmarks), especially for larger queries that return payloads above > 10MB. This issue https://github.com/colinhacks/zod/issues/205 confirms others are hitting this with zod as well.

We don't really need to parse zoekt's response bodies (since we always expect them to be valid), so this PR removes the `parseAsync` call and instead does a simple cast. The results are pretty dramatic: my initial benchmarks are shoing a increase in search performance by a order of magnitude (89.8% reduction in average search time). Here's the before & after on the benchmark:
```
Before:
📈 Latency Distribution:
   Min (p0):    407.04ms
   p50:        2660.49ms
   p75:       11770.50ms
   p90:       50848.31ms
   p95:       54193.05ms
   p99:       74902.95ms
   Max (p100):84267.76ms

After:
📈 Latency Distribution:
   Min (p0):    293.31ms
   p50:         569.53ms
   p75:        1855.31ms
   p90:        3682.20ms
   p95:        4733.21ms
   p99:        5561.01ms
   Max (p100): 5805.22ms
```

With this change, I figured we can bump the default number of results requested to 100k (up from 5k!). In my testing performance was good and I was able to get 100k results in <5 seconds.

## Summary by CodeRabbit

* **New Features**
  * Added debug timing information to search responses for performance visibility.

* **Changed**
  * Increased default search result count from 5,000 to 100,000.

* **Bug Fixes**
  * Resolved a significant performance bottleneck in the search API, yielding substantial performance gains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed race condition in job schedulers. [#607](https://github.com/sourcebot-dev/sourcebot/pull/607)
 - Fixed connection sync jobs getting stuck in pending or in progress after restarting the worker. [#612](https://github.com/sourcebot-dev/sourcebot/pull/612)
 - Fixed issue where connections would always sync on startup, regardless if they changed or not. [#613](https://github.com/sourcebot-dev/sourcebot/pull/613)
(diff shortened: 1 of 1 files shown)


brendan-kellam:
@coderabbitai review

PR #617, opened 2025-11-14
fix(web): Fix loading issues with references / definitions list

brendan-kellam:
We were getting spurious loading issues with the references / definitions panel and file tree. The root cause of this was we were using server actions which cannot run in parallel and are ill-suited for client side queries. This PR migrates a couple of actions over to REST endpoints.

## Release Notes

* **Bug Fixes**
  * Fixed spurious infinite loads occurring in the explore panel, file tree, and file search command.


brendan-kellam:
@coderabbitai review

PR #642, opened 2025-11-29
Shrink Docker image size by ~1/3 by removing unnecessary ops

thespad:
Current image does two very inefficient operations:

* It copies node_modules from `shared-libs-builder` stage to `backend-builder` stage and then in the `runner` stage copies node_modules from both stages again, even though the `backend-builder` stage now contains all the necessary modules.
* It chowns the whole of /app *after* COPYing the files from previous stages

Both of these result in duplicating all of the affected files in the `runner` stage layers, which gives you about 600Mb of wasted space.

By removing the duplicate COPY of node_modules, and performing the chown as part of the COPY operation, that space can be free, reducing the image size substantially.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -250,8 +250,10 @@ RUN mkdir -p /run/postgresql && \
 RUN chown -R sourcebot:sourcebot /data
 
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
-COPY --chmod=+x prefix-output.sh ./prefix-output.sh
-COPY --chmod=+x entrypoint.sh ./entrypoint.sh
+COPY prefix-output.sh ./prefix-output.sh
+RUN chmod +x ./prefix-output.sh
+COPY entrypoint.sh ./entrypoint.sh
+RUN chmod +x ./entrypoint.sh
 
 # Note: for back-compat cases, we do _not_ set the USER directive here.
 # Instead, the user can be overridden at runtime with --user flag.


brendan-kellam:
Thanks for the PR this is really great!! Heads up, I was hitting the following error when building locally:
```
ERROR: failed to solve: invalid chmod parameter: '+x'. it should be octal string and between 0 and 07777
```

It's probably because whatever bundled Dockerfile syntax version on my machine is less than [version 1.14.0](https://docs.docker.com/build/buildkit/dockerfile-release-notes/#1140) where they introduced non-octal chmod params.

I think it would be best to add `# syntax=docker/dockerfile:1` to the top of the Dockerfile such that we use the latest Dockerfile syntax. ([docs](https://docs.docker.com/reference/dockerfile/#syntax))

thespad:
Done. It's what I'd normally do when building a Dockerfile from scratch but I try not to assume too much when modifying other peoples' work.

Dockerfile
```diff
@@ -245,16 +247,11 @@ RUN mkdir -p /run/postgresql && \
...
-# Make data directory accessible to both root and sourcebot user
 RUN chown -R sourcebot:sourcebot /data
 
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
-COPY prefix-output.sh ./prefix-output.sh
-RUN chmod +x ./prefix-output.sh
-COPY entrypoint.sh ./entrypoint.sh
-RUN chmod +x ./entrypoint.sh
+COPY --chmod=+x prefix-output.sh ./prefix-output.sh
+COPY --chmod=+x entrypoint.sh ./entrypoint.sh
```
brendan-kellam: Hitting "sourcebot   | [FATAL tini (7)] exec ./entrypoint.sh failed: Permission denied" when running the container as non-root (`--user sourcebot`). These probably need --chown operators too.

thespad: Odd, I can't replicate with a local build
```
source  | [Info] Running as non-root user.
source  | [Info] Resolving environment overrides from /data/config.json...
source  | [Info] Sourcebot version: 
source  | [Warning] NEXT_PUBLIC_POSTHOG_PAPIK was not set. Setting SOURCEBOT_TELEMETRY_DISABLED.
source  | [Info] Disabling telemetry since SOURCEBOT_TELEMETRY_DISABLED was set.
source  | [Info] Running database migration...
source  | Prisma schema loaded from prisma/schema.prisma
source  | Datasource "db": PostgreSQL database "postgres", schema "public" at "source-postgres:5432"
source  | 
source  | 46 migrations found in prisma/migrations
source  | 
source  | Applying migration `20250122225856_postgres_init`
```
That code hasn't fundamentally changed, the original was a COPY then a chmod +x and the new is just a COPY *with* a chmod +x

brendan-kellam: hmm so looks like if I build with the separate COPY and RUN commands, things work correctly. I asked gpt and it seems like I don't have Docker buildkit enabled. I tried building with `DOCKER_BUILDKIT=1` but that didn't solve the problem.

I'll keep diagnosing on my side, but I'm leaning to having these be separate commands like before to avoid situations where people run into issues like I'm seeing on my machine.

thespad: Buildkit has been the default builder for a while now, so you shouldn't have to enable it unless you're on a rather old version, but there's no real issue reverting those lines as the entrypoint files are tiny so it's more a point of consistency than actually saving a lot of space.

brendan-kellam: Yea I'm not totally sure what is going on with it - I'm on a recent build of docker desktop and it does look like buildkit is enabled. This is probably a _me_ problem, but considering that it won't impact the size of the image I'd say let's just revert to what it was.

thespad: Fixed in [`86219a9` (#642)](https://github.com/sourcebot-dev/sourcebot/pull/642/commits/86219a92ba93d1c7d5aa4efc1dfb7d84ec6e5bd8)

brendan-kellam (approved):
thank you for this PR, it's super helpful!

PR #647, opened 2025-12-01
chore(web): Scope code nav to current repository by default

brendan-kellam:
We were seeing that on large deployments, code navigation queries were sometimes taking several seconds to resolve. This was because we were searching across all repositories to resolve references / definitions.

This PR changes the default behaviour to only search in the current repository where the user is. I've also added a toggle button to allow searching across all repositories as well:

image

The rationale / intuition here is that, most of the time, symbol are only referenced within the repository they are defined in. There are exceptions of course (e.g., methods exported from a shared library), but most of the time this will be the case. Besides speed, the other benefit of scoping to the current repository as it reduces the false-positive rate inherent with search-based code navigation (e.g., if function `doSomething` is declared in multiple function).

## Summary by CodeRabbit

* **New Features**
  * Added cross-repository search toggle in the explore menu with Shift+A keyboard shortcut support.
  * Code navigation now displays keyboard shortcut hints for improved discoverability.

* **Bug Fixes**
  * Fixed search query handling for repository names containing special characters.

* **Documentation**
  * Updated feature documentation to reflect cross-repository navigation capabilities.

* **Chores**
  * Enhanced analytics event tracking for code navigation and search interactions.

### DIFF
--- a/packages/web/src/components/ui/toggle.tsx
+++ b/packages/web/src/components/ui/toggle.tsx
@@ -17,8 +17,6 @@ const toggleVariants = cva(
       },
       size: {
         default: "h-7 w-7 min-w-7 p-0",
-        sm: "h-9 px-2.5 min-w-9",
-        lg: "h-11 px-5 min-w-11",
       },
     },
     defaultVariants: {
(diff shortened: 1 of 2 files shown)


brendan-kellam:
@coderabbitai review

PR #652, opened 2025-12-03
chore(web): Server side search telemetry

brendan-kellam:
This PR adds server side telemetry for search

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


packages/web/src/lib/posthogEvents.ts
```diff
@@ -313,6 +313,11 @@ export type PosthogEventMap = {
         durationMs: number,
         // Whether or not the user is searching all repositories.
         isGlobalSearchEnabled: boolean,
-    }
+    },
+    //////////////////////////////////////////////////////////////////
+    api_code_search_request: {
```
brendan-kellam, line 318: We should instrument our other api endpoints too

PR #658, opened 2025-12-04
fix: add support for anyuid to Dockerfile

Cschlaefli:
This change updates the Containerfile to [support arbitrary userids](https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/html/creating_images/creating-images-guidelines#openshift-specific-guidelines) to enable running in Openshift.

It does this by adding group write permissions to `/app` and setting both `/app` and `/data` to use the root group by default. 

This should not have any impact on existing environments running as `sourcebot` user since all of the directories changed are owned by the `sourcebot` user.

## Summary by CodeRabbit

* **Chores**
  * Updated container directory access permissions to enable multi-user group-based access on the application directory, while maintaining sourcebot-specific ownership on the data directory.


brendan-kellam:
@coderabbitai review

brendan-kellam (approved):

PR #739, opened 2026-01-15
fix: Improve initial file tree load performance

### LINKED ISSUE #651: [bug] File tree can take a long time to load
### Describe the bug

For larger repositories, the file tree can take upwards of 10-20 seconds to load.

### To reproduce

Open a larger project in browse.

Sourcebot version (e.g. v3.0.1): latest

### Additional information

My guess is that the bulk of the time is in the `ls-tree` command, as well as transmitting the large json blob. We could probably only load in a subset of the information, and then dynamically request additional data as the user expands folders.

brendan-kellam:
## Problem
For large repositories (100K+ files), the file tree could take a long time (10s+) to load. The primary bottleneck was because we were loading the entire file tree for the repository, even though a small subset would actually be rendered.

## Solution
This PR improves things by only loading the **subset of the file tree** that is actually opened*, rather than loading everything. The tradeoff with this approach is that subsequent requests need to be made whenever a new folder is opened. When this happens, a "loading..." placeholder is rendered below the folder. Upon testing on a production build, this experience is still pretty snappy (was seeing 50ms load times). I've also added a `wa_file_tree_loaded` posthog event that includes a `durationMs` so we can track the performance in the wild.

*opened in this context doesn't necessarily mean rendered. If you open a nested directory and then collapse some parent directory, the nested directory will still be considered open but not rendered. This is needed since if you were to then expand the parent directory, the state of the opened directories should still be preserved.

### Before

https://github.com/user-attachments/assets/28a11b05-4e35-4767-8bf7-81b1920169a8

### After

https://github.com/user-attachments/assets/cfedb1ae-a3e2-4da1-9032-5315504d1ac1

Fixes #651 

---

> [!NOTE]
> Speeds up initial file tree rendering by fetching only opened directories and deferring the rest; adds basic telemetry to measure load times.
> 
> - Change `getTree` API to accept `paths` and return a union tree; add `getFolderContents`, plus `normalizePath`/`isPathValid`/`buildFileTree` helpers and `compareFileTreeItems` sorter (with tests)
> - Update UI: `FileTreePanel` tracks `openPaths`, queries tree on expand (React Query with `staleTime: Infinity`, placeholder data), emits `wa_file_tree_loaded` with `durationMs`; `PureFileTreePanel` uses `openPaths` and shows a loading skeleton for unopened/empty children
> - Add PostHog event `wa_file_tree_loaded`; update types/schemas accordingly; note performance improvement in `CHANGELOG.md`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 279e04435e9dc4478468823032f2370059d29785. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Faster file tree loading for large repositories with optimized path handling
  * Added loading skeleton feedback when expanding directories

* **Bug Fixes**
  * Fixed error when viewing directories with special path syntax (..)

* **Documentation**
  * Updated changelog with performance improvements and bug fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Improved initial file tree load times, especially for larger repositories. [#739](https://github.com/sourcebot-dev/sourcebot/pull/739)
 
-### Fixed
-- Fixed "error loading tree preview" error when viewing a directory with `...`. [#531](https://github.com/sourcebot-dev/sourcebot/issues/531)
-
 ## [4.10.9] - 2026-01-14
 
 ### Changed


packages/web/src/app/api/(client)/client.ts
```diff
@@ -101,4 +103,12 @@ export const getFiles = async (body: GetFilesRequest): Promise<GetFilesResponse
         body: JSON.stringify(body),
     }).then(response => response.json());
     return result as GetFilesResponse | ServiceError;
+}
+
+export const getFolderContents = async (body: GetFolderContentsRequest): Promise<GetFolderContentsResponse | ServiceError> => {
```
brendan-kellam: no longer needed

packages/web/src/features/fileTree/components/fileTreePanel.tsx
```diff
@@ -41,17 +40,79 @@ export const FileTreePanel = ({ order }: FileTreePanelProps) => {
...
-        queryKey: ['tree', repoName, revisionName],
-        queryFn: () => unwrapServiceError(
-            getTree({
-                repoName,
-                revisionName: revisionName ?? 'HEAD',
-            })
-        ),
+
+    const { data, isError } = useQuery({
+        queryKey: ['tree', repoName, revisionName, ...Array.from(openPaths)],
```
brendan-kellam, line 49: todo: seems like we are still re-fetching even though the key should be stable

packages/web/src/features/fileTree/components/pureFileTreePanel.tsx
```diff
@@ -110,13 +68,15 @@ export const PureFileTreePanel = ({ tree: _tree, path }: PureFileTreePanelProps)
                                 }}
                                 parentRef={scrollAreaRef}
                             />
-                            {node.children.length > 0 && !node.isCollapsed && renderTree(node, depth + 1)}
+                            {node.type === 'tree' && node.children.length > 0 && openPaths.has(node.path) && renderTree(node, depth + 1)}
+                            {/* @note: a empty tree indicates that the contents are beaing loaded. Render a loading skeleton to indicate that. */}
+                            {node.type === 'tree' && node.children.length === 0 && openPaths.has(node.path) && renderLoadingSkeleton(depth)}
```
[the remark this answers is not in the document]
brendan-kellam: You cannot have a empty path in git.

brendan-kellam:
@coderabbitai review

Issue #661, opened 2025-12-05
[FR] Allow anonymous user to sign in to save their chat


msukkari:
Chats from anonymous users are saved in the DB but not associated with a user, and anonymous users can't view their chat history. Enable an anonymous user to log in while in a chat session to enable them to save the chat into their newly created user.

Pallavikumarimdb:
Hi! @brendan-kellam @msukkari , I’d like to work on this issue and wanted to share a proposed approach.
After looking through the current auth and anonymous-access flow, anonymous access is supported at the org level.

Proposed approach:

1. Anonymous session identity
- Generate a stable anonymous session ID (UUID v4) on first load for unauthenticated users.
- Store it in a cookie so it persists across refreshes and tabs within the same browser.
- Use this ID to associate chats created while the user is anonymous.

2. Chat ownership (minimal schema change)
- Since `Chat.createdById` is nullable and there is currently no alternative
  identifier for anonymous users, introduce an optional field on `Chat`
  (e.g. `anonSessionId`) to temporarily associate chats with an anonymous session.
- Authenticated chats continue to use `createdById` as-is.

3. Login / signup migration
- On successful login or signup, read the anonymous session ID from the cookie.
- Migrate all chats associated with that ID to the authenticated user
  (assign `createdById`, clear the anonymous session ID).
- Clear the anonymous session cookie after migration.
- Migration is scoped to the current browser/session and is idempotent.

Would appreciate your feedback on this approach before I start!
Happy to discuss any concerns or alternative approaches.

brendan-kellam:
Hey @Pallavikumarimdb thanks for the input! I think this approach makes sense. Where do you foresee the migration taking place?

Pallavikumarimdb:
Hi @brendan-kellam I’d expect the migration to happen server-side immediately after a successful login or signup, once we have a concrete `userId`.

- Read the anonymous session ID from the request cookies.
- In the same request lifecycle (or a small follow-up action), migrate any chats associated with that anonymous ID by assigning `createdById` and clearing the anonymous ID.
- Then clear the anonymous session cookie.

msukkari:
Not sure if relevant here but something to note here is how this plays with the anonymous user ID generated with/for posthog. Would be great if these were correlated somehow

PR #740, opened 2026-01-15
feat: Allow anonymous user to sign in to save their chat

Pallavikumarimdb:
Fixes: #661 

This PR introduces:
* When anonymous user sign in or create an account, their anonymous chats are automatically migrated to their authenticated account.


brendan-kellam:
Closing since this was implemented in #888

PR #754, opened 2026-01-18
feat(web): Share chat session link publicly 

### LINKED ISSUE #479: [FR] Share links for Ask Sourcebot chat sessions
I found that after logging in, the Chat Session can only be accessed by the logged-in user and cannot be changed to Public.

Nil2000:
Fixes #479 
For now I have added a simple dialog and action to change the visibility.

Need confirmation if approach is correct or not 

Preview below
[Screencast From 2026-01-18 21-15-09.webm](https://github.com/user-attachments/assets/1ad6a728-f2bc-4614-b31f-1f43220a87fc)

After this, I was thinking of implementing a publicly available page `/share/chat/<chatid>`
It will reuse chat page components but will not use `<Chatbox/>`. Will keep Read-only by default based on visibility.

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -74,6 +74,7 @@
     "@radix-ui/react-label": "^2.1.0",
     "@radix-ui/react-navigation-menu": "^1.2.0",
     "@radix-ui/react-popover": "^1.1.6",
+    "@radix-ui/react-radio-group": "^1.3.8",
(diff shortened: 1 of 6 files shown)


Nil2000:
@brendan-kellam @msukkari Would be great if you can check my approach once.

brendan-kellam:
Thanks for the PR, but ended up implemented this in #888

Issue #679, opened 2025-12-17
[FR] Dynamic GitLab project selection via API queries


philippe-granet:
### Description

Add support for configuring which GitLab projects are mirrored/synchronized by using GitLab API queries.

Introduce a new optional parameter called `projectQuery`, which allows users to dynamically select projects to mirror based on GitLab API endpoints.

### Proposed Configuration

The `projectQuery` parameter would be an array of strings.  
Each string represents a GitLab API URL path (including query parameters) that returns a list of projects.

Example:

```json
{
  "connections": {
    "gitlab-connection": {
      "type": "gitlab",
      "groups": [
        "..."
      ],
      "projects": [
        "..."
      ],
      "projectQuery": [
        "groups/group1/projects?include_subgroups=true",
        "groups/group2/projects?include_subgroups=false",
        "groups/group3/projects?include_subgroups=true&archived=false",
        "search?scope=projects&search=flight",
        "projects?topic=devops"
      ]
    }
  }
}
```

Here are a few endpoints that return a list of projects:
* https://docs.gitlab.com/api/projects.html#list-all-projects
* https://docs.gitlab.com/api/groups.html#list-a-groups-projects
* https://docs.gitlab.com/api/search.html#scope-projects.

### Expected Behavior

All projects returned by the specified API queries should be mirrored in Sourcebot.
If a project is matched by multiple query strings, it must be imported only once (deduplication).
This feature should complement existing groups and projects configuration options.

### Motivation
This would make GitLab synchronization more flexible and scalable, especially for organizations with:
* Large numbers of projects
*  Complex group and subgroup structures

It would also reduce configuration maintenance by relying on GitLab’s API filtering capabilities.

sudhanshu112233shukla:
do check the PR i have made some feature !!
tested it as well !!!

GitBalake:
archived=false

This would be nice to have as a top level group setting as well. 

So subrepos in groups are naturally excluded if archived.

PR #757, opened 2026-01-20
feat(gitlab): add projectQuery support for dynamic project discovery

sudhanshu112233shukla:
Fixes:#679
This PR addresses issue #679 by adding support for a new projectQuery configuration options in the GitLab integration. This allows users to dynamically select projects to mirror using raw GitLab API query strings, offering greater flexibility than statically defining groups or projects.
**Approach**
**Schema Update:**
**Modified :**
schemas/v3/gitlab.json to include the projectQuery property (an array of strings).
Updated the description and added examples to guide users on how to construct API query paths (e.g., groups/my-group/projects?include_subgroups=true).

**Backend Implementation:**
Updated 
packages/backend/src/gitlab.ts
 to process the projectQuery array.
**Implemented** 
- parseQuery helper to convert query strings into a format compatible with the gitbeaker library (handling booleans and numbers correctly).
- Used project.Projects.all({...}) to fetch projects matching the parsed queries.
- Deduplication: Implemented logic to ensure that if a project is returned by multiple queries (or overlaps with other config options), it is only added to the list once based on its Project ID.

**Deduplication**: Implemented logic to ensure that if a project is returned by multiple queries (or overlaps with other config options), it is only added to the list once based on its Project ID.

**Testing**
I added comprehensive tests in packages/backend/src/gitlab.test.ts to fully validate the new functionality. This includes unit tests to ensure the parseQuery helper correctly converts query strings into the expected object parameters. I also added mocked integration tests by stubbing the gitbeaker GitLab client and the api.Projects.all method, verifying that getGitLabReposFromConfig invokes the GitLab API with parameters accurately derived from the configuration strings. The tests confirm that results returned from multiple queries are properly merged and deduplicated, while also ensuring that existing exclusion logic (such as exclude.topics) continues to work as intended. All tests were successfully verified by running yarn workspace @sourcebot/backend test src/gitlab.test.ts.

### DIFF
--- a/packages/backend/src/gitlab.test.ts
+++ b/packages/backend/src/gitlab.test.ts
@@ -1,5 +1,52 @@
-import { expect, test } from 'vitest';
-import { shouldExcludeProject } from './gitlab';
+import { expect, test, vi, describe, beforeEach } from 'vitest';
+import { shouldExcludeProject, parseQuery, getGitLabReposFromConfig } from './gitlab';
+import { ProjectSchema } from '@gitbeaker/rest';
+
+// Mock dependencies
+const mockGitlabAll = vi.fn();
+
+vi.mock('@gitbeaker/rest', () => {
+    return {
+        Gitlab: vi.fn().mockImplementation(() => ({
+            Projects: {
+                all: mockGitlabAll,
+            }
+        })),
+    };
+});
+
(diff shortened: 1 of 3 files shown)


sudhanshu112233shukla:
@msukkari @brendan-kellam do check it !!

Issue #768, opened 2026-01-21
[bug] The data directory was initialized by PostgreSQL version 16, which is not compatible with this version 18.1.


CyberT17:
### Describe the bug

After upgrading to v4.10.3, the upgrade fails because the previous version (v4.10.12) was using Postgres version 16 and the new node image installs postgres version 18.

```
2026-01-21 15:22:52.902 UTC [79] FATAL:  database files are incompatible with server
2026-01-21 15:22:52.902 UTC [79] DETAIL:  The data directory was initialized by PostgreSQL version 16, which is not compatible with this version 18.1.
postgres failed to run
```

### To reproduce

Upgrade and deploy to v4.10.13 with an already populated database. 

### Sourcebot deployment information

Sourcebot version (e.g. v3.0.1):

sudhanshu112233shukla:
this PR will fix your issue 

thank you!!

brendan-kellam:
@CyberT17 are you using the docker-compose or the database embedded in the Dockerfile?

CyberT17:
db embedded in the Dockerfile

PR #769, opened 2026-01-21
fix: pin postgres version to 16 to avoid data incompatibility

sudhanshu112233shukla:
fixes:#768

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ARG NEXT_PUBLIC_SENTRY_BACKEND_DSN
 ARG NEXT_PUBLIC_LANGFUSE_PUBLIC_KEY
 ARG NEXT_PUBLIC_LANGFUSE_BASE_URL
 
-FROM node:24-alpine3.23 AS node-alpine
+FROM node:20-alpine3.19 AS node-alpine
 FROM golang:1.23.4-alpine3.19 AS go-alpine
 # ----------------------------------
 


CyberT17:
I don't believe this fixes the issue. The issue is in how the docker image is created using the dockerfile.

sudhanshu112233shukla:
@CyberT17 Thanks for pointing this out  you were absolutely right. The root cause was the Dockerfile using a newer base image (node:24-alpine3.23), which implicitly pulled in Postgres 18 and led to the crash. I’ve now updated the PR to address this thoroughly. The Dockerfile has been downgraded to node:20-alpine3.19, ensuring that the embedded Postgres binary is version 16, which matches the Alpine 3.19 packages and resolves the issue for single-container deployments. In addition, the docker-compose.yml now explicitly pins POSTGRES_VERSION to 16, guaranteeing the same behavior and consistency for Compose users. With these changes in place, the upgrade incompatibility should be fully resolved and the PR is ready for re-review @msukkari @brendan-kellam

brendan-kellam:
> @CyberT17 Thanks for pointing this out you were absolutely right. The root cause was the Dockerfile using a newer base image (node:24-alpine3.23), which implicitly pulled in Postgres 18 and led to the crash. I’ve now updated the PR to address this thoroughly. The Dockerfile has been downgraded to node:20-alpine3.19, ensuring that the embedded Postgres binary is version 16, which matches the Alpine 3.19 packages and resolves the issue for single-container deployments. In addition, the docker-compose.yml now explicitly pins POSTGRES_VERSION to 16, guaranteeing the same behavior and consistency for Compose users. With these changes in place, the upgrade incompatibility should be fully resolved and the PR is ready for re-review @msukkari @brendan-kellam

alpine 3.23 does appear to implicitly pull in postgres 18, but the fix shouldn't be to downgrade back to node:20-alpine3.19. Instead, pining the postgres version explicitly to 17 is a better solution. I'll open up a PR that does this.

sudhanshu112233shukla:
@brendan-kellam got it !

Issue #771, opened 2026-01-21
[bug] Ask cannot handle files with special characters


marshallwp:
### Describe the bug

Say you have a file named: `(pr` that you want to find. You can search for this with Sourcebot via the query `file:"\(pr"`.

If you Ask Sourcebot about it, it will try a number of things, all of which either result in query failure or returning incorrect results.

Query: `find me file:"\(pr"`

<details><summary>GPT-5-nano Response</summary>

1. Searched for `\(pr`
Failed with the following error: `Failed to parse query "\(pr reposet:gitlab/example-repo" with message: No parse at 65`

2. Searched for `\(pr ( repo:gitlab/example-repo )`
Failed with the following error: `Failed to parse query "\(pr reposet:gitlab/example-repo ( repo:gitlab/example-repo )" with message: No parse at 127`

3. Searched for `pr ( repo:gitlab/example-repo )`
Returned src/app/routes/endpoint.py
</details>

<details><summary>GPT-5-mini Response</summary>

1. I'll search the repository for files containing the literal substring "(pr" using the code search tool to locate all matches. This helps find files that include that pattern (likely a regex or string with an escaped parenthesis). I'll search only the selected repository.
Searched for `\\(pr ( repo:gitlab/example-repo )`
Failed with the following error: `Failed to parse query "\\(pr reposet:gitlab/example-repo ( repo:gitlab/example-repo )" with message: No parse at 128`

2. Searched for `"\\(pr" ( repo:gitlab/example-repo )`
Failed with the following error: `Unexpected error: An unexpected error occurred. Please try again later.`
</details>

<details><summary>Anthropic Claude Haiku 4.5</summary>

1. The user is asking me to find files with a pattern "\(pr". This seems to be a search for code that contains this literal string pattern - likely a regex or string pattern in code.<br><br>Let me use the searchCode function to find this pattern. The user wants to search for the literal string "\(pr" which appears to be a regex or escape sequence pattern.<br><br>Since this is a regex search, I should search for files that contain this pattern. The user's query is "\(pr" which I should escape properly for a regex search.<br><br>Let me search for this code pattern.<br>"Search code" tool call failed

2. Searched for `\\"\(pr ( repo:example-repo )`
Failed with the following error: `Failed to parse query "\\"\(pr reposet:gitlab/example-repo ( repo:example-repo )" with message: No parse at 90`

3. Let me try a simpler regex pattern:
Searched for `\(pr ( repo:example-repo )`
[...]

PR #786, opened 2026-01-23
fix(ask): Properly handle regex filters that include parenthesis

msukkari:
Weird edge case where if a regex filter like filename/reponame contained a parenthesis the parser would fail. This is because the query language forbids parenthesis in the value since they're used to match against the ParenExpr symbol.

The ideal case is that the agent would be smart enough to wrap the regexp in quotes, which would solve this problem. I've added instructions in the tool call to tell it to do this, but it rarely does. As a result, I've added basic logic to process the regexp before adding it in the query string to wrap it in quotes if a parenthesis is included.

Fixes https://github.com/sourcebot-dev/sourcebot/issues/771

## Summary by CodeRabbit

* **Bug Fixes**
  * Properly handle regex filters that include parentheses.

* **New Features**
  * Added file path filtering capability to code search functionality.

* **Documentation**
  * Updated guidance on using file filters with parentheses in regex expressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Properly map all hotkeys in UI based on the platform [#784](https://github.com/sourcebot-dev/sourcebot/pull/784)
+- Properly map all hotkeys in UI based on the platform. [#784](https://github.com/sourcebot-dev/sourcebot/pull/784)
+- Properly handle regex filters that include parenthesis. [#786](https://github.com/sourcebot-dev/sourcebot/pull/786)
 
 ## [4.10.16] - 2026-01-22
 


packages/mcp/src/index.ts
```diff
@@ -3,6 +3,7 @@
 // Entry point for the MCP server
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { preprocessRegexp } from '@sourcebot/shared';
```
brendan-kellam, line 6: nit: mcp cannot take a dependency on `@sourcebot/shared` since we do not publish it to npm

brendan-kellam:
I wonder if we could improve our parsing to not throw a exception in the case of a unmatched paren?

msukkari:
> I wonder if we could improve our parsing to not throw a exception in the case of a unmatched paren?

Closing in favor of https://github.com/sourcebot-dev/sourcebot/pull/788

PR #805, opened 2026-01-28
fix(worker): Add tmp shard cleanup on indexing failure

### LINKED ISSUE #804: [bug] Left over indexing shards are filling up the disk slowly
We have about 300gb allocated to sourcebot, and normally it only takes up 100gb of storage. But we have noticed that there are a lot of .tmp zoekt shards that are left behind that slowly take over more and more space. 
The ID on these temp files matches with the repo ID that constantly fail to index.
Here is a chart to show you what we are seeing. 
We have created a clean up script that we have to run from time to time. 

We have a bunch of large repos that zoekt has never been able to index. Which I suspect might be causing this?

Image

Here is an error log that I frequently see. (Also quick aside, whenever the zoekt indexer throws an error the node process gets killed, which cause the next app to restart. Minor inconvenience)

```
stack:  Error: Command failed: zoekt-git-index -allow_missing_branches -index /data/.sourcebot/index -max_trigram_count 50000 -file_limit 2097152 -branches "HEAD" -tenant_id 1 -repo_id 1618 -shard_prefix 1_1618 -large_file "..." /data/.sourcebot/repos/1618 
2026/01/28 16:11:14 attempting to index 6560 total files 
2026/01/28 16:11:28 symbol analysis finished for shard statistics: duration=1s symbols=1 bytes=105038110 
at genericNodeError (node:internal/errors:985:15) 
at wrappedFn (node:internal/errors:539:14) 
at ChildProcess.exithandler (node:child_process:417:12) 
at ChildProcess.emit (node:events:508:28) 
at maybeClose (node:internal/child_process:1101:16) 
at Socket.<anonymous> (node:internal/child_process:457:11) 
at Socket.emit (node:events:508:28) 
at Pipe.<anonymous> (node:net:346:12)

```

msukkari:
Clean up temporary Zoekt shard files on indexing failure to prevent disk space exhaustion.

When `zoekt-git-index` fails during repository indexing, it leaves behind `.tmp` shard files. These accumulate over time, especially for repos that repeatedly fail to index, leading to disk space issues. This PR adds logic to automatically remove these temporary files immediately after an indexing operation fails.

Fixes #804

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed issue where temporary files created during index operations were not being cleaned up when failures occurred, which could accumulate and consume unnecessary disk space.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added support to set "Require approval for new members" via config with (`REQUIRE_APPROVAL_NEW_MEMBERS`). [#858](https://github.com/sourcebot-dev/sourcebot/pull/858)
-- Added automatic cleanup of tmp shard files on indexing failure. [#805](https://github.com/sourcebot-dev/sourcebot/pull/805)
 
 ### Changed
 - Improved stability for connection and repo indexing workers. [#860](https://github.com/sourcebot-dev/sourcebot/pull/860)
@@ -17,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed issue where certain file and folder names would cause type errors. [#862](https://github.com/sourcebot-dev/sourcebot/pull/862)
 - Fixed token refresh error "Provider config not found or invalid for: x" when a sso is configured using deprecated env vars. [#841](https://github.com/sourcebot-dev/sourcebot/pull/841)
+- Fixed issue where temporary shard files created on index failure were not being cleaned up. [#805](https://github.com/sourcebot-dev/sourcebot/pull/805)
 
 ## [4.10.27] - 2026-02-05
 


packages/backend/src/repoIndexManager.ts
```diff
@@ -438,9 +438,17 @@ export class RepoIndexManager {
         }
 
         logger.info(`Indexing ${repo.name} (id: ${repo.id})...`);
-        const { durationMs } = await measure(() => indexGitRepository(repo, this.settings, revisions, signal));
-        const indexDuration_s = durationMs / 1000;
-        logger.info(`Indexed ${repo.name} (id: ${repo.id}) in ${indexDuration_s}s`);
+        try {
+            const { durationMs } = await measure(() => indexGitRepository(repo, this.settings, revisions, signal));
+            const indexDuration_s = durationMs / 1000;
+            logger.info(`Indexed ${repo.name} (id: ${repo.id}) in ${indexDuration_s}s`);
```
brendan-kellam, line 483: I think this is a fair workaround and we should probably have a mechanism for cleaning up these files, but it doesn't really address the root cause for why these files are being created in the first place.

PR #806, opened 2026-01-28
[SOU-301] Fix branch filtering to use short names instead of refs/heads

### LINKED ISSUE #800: Searching on filtered branches returns no results
When I go into the repository and try to filter on indexed branches, no results are returned. 

Image

Image

msukkari:
## Problem
Searching on filtered branches was returning no results because branch names were being stored with the `refs/heads/` prefix when they should have been stored as short names only.

## Solution
Remove the unnecessary mapping that prefixed branch names with `refs/heads/` in the branch filtering logic. The zoekt index stores branch references as short names (e.g., `main`, `develop`), so the filtering logic should match this format.

## Changes
- Removed `.map((branch) => \`refs/heads/${branch}\`)` from branch filtering in `repoIndexManager.ts`
- Branch names are now correctly passed as short names to match indexed branch references

This fixes the issue where the `rev:` query syntax was not finding any results when filtering on indexed branches.

Fixes #800 

---
[View Niteshift Task](https://niteshift.dev/repo/sourcebot-dev__sourcebot/task_ubfqugon-r_l0cjb3azzl)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository branch selection processing for indexing operations.

### DIFF
--- a/packages/backend/src/repoIndexManager.ts
+++ b/packages/backend/src/repoIndexManager.ts
@@ -404,8 +404,7 @@ export class RepoIndexManager {
             const allBranches = await getBranches(repoPath);
             const matchingBranches =
                 allBranches
-                    .filter((branch) => micromatch.isMatch(branch, branchGlobs))
-                    .map((branch) => `refs/heads/${branch}`);
+                    .filter((branch) => micromatch.isMatch(branch, branchGlobs));
 
             revisions = [
                 ...revisions,


packages/backend/src/repoIndexManager.ts
```diff
@@ -404,8 +404,7 @@ export class RepoIndexManager {
             const allBranches = await getBranches(repoPath);
             const matchingBranches =
                 allBranches
-                    .filter((branch) => micromatch.isMatch(branch, branchGlobs))
-                    .map((branch) => `refs/heads/${branch}`);
+                    .filter((branch) => micromatch.isMatch(branch, branchGlobs));
```
brendan-kellam, line 407: I don't think this is a correct change. #800 is referring to the filter on the `/repos` page. We need to have `refs/heads/` and `refs/tags/` to allow the user to distinguish between filtering a branch and a tag (e.g., in the event where a tag and branch are named the same thing). This is documented here: https://docs.sourcebot.dev/docs/features/search/multi-branch-indexing#search-syntax

CyberT17: The current suggested change would still be a welcomed change. 
1. Making sure `rev:` maps to the short name of the default branch makes it easier for non technical people to understand which branch they are looking at. They might not know what `refs/heads` stands for, but they would know what `main`, `feature/`, etc are as that is generally what is displayed in all code hosting platforms
2. If you specify, which branches to index in the SB configs, like this 
```      
    "revisions": {
        "branches":
          "main"
          "release/*"
        ]
```
it indexes main and ref/head separately (iirc)
3. I am working on adding the feature to be able to switch revisions directly from the code tree/preview panels. and this would be useful in my efforts. (and goes hand in hand with 1. above)

brendan-kellam: > 1. Making sure rev: maps to the short name of the default branch makes it easier for non technical people to understand which branch they are looking at.

I believe zoekt does a substring match. For example:
image
image

I would like to keep it such that a user can explicitly specify a `refs/heads` or `refs/tags` like this. In the UI, I think we can just strip the refs/heads & refs/tags

> 2.If you specify, which branches to index in the SB configs, like this it indexes main and ref/head separately (iirc)

This actually appears to be a bug. Tracked in #808 

> 3. I am working on adding the feature to be able to switch revisions directly from the code tree/preview panels. and this would be useful in my efforts. (and goes hand in hand with 1. above)

Cool! I assume you mean something like a dropdown selector that lets users switch between branches? Created #810 to track this if you want to discuss over there.

brendan-kellam:
Fixed in #851

PR #870, opened 2026-02-10
feat(mcp): add list_tree tool

brianphillips:
This PR adds a `list_tree` tool to the MCP package. This tool allows an agent to list files and directories from a repository path. This can be used as a simple directory listing tool or for a directory tree tool by specifying a depth > 1

## Summary by CodeRabbit

* **New Features**
  * Added a new list_tree tool to browse repository trees with depth control, includeFiles/includeDirectories toggles, and a maxEntries limit.

* **Documentation**
  * Added usage details, parameters, and examples for the new list_tree tool in docs, README, and changelog.

* **Bug Fixes**
  * Improved repository-not-found message to include the repository name for clearer diagnostics.

### DIFF
--- a/packages/web/src/features/git/getTreeApi.ts
+++ b/packages/web/src/features/git/getTreeApi.ts
@@ -34,7 +34,7 @@ export const getTree = async ({ repoName, revisionName, paths }: GetTreeRequest)
         });
 
         if (!repo) {
-            return notFound();
+            return notFound(`Repository "${repoName}" not found.`);
         }
 
         const { path: repoPath } = getRepoPath(repo);


packages/mcp/src/index.ts
```diff
@@ -238,6 +239,155 @@ server.tool(
     }
 );
 
+server.tool(
```
brendan-kellam, line 242: This approach makes sense for the apis that we expose, and I'm happy to ship this. For context,`/api/tree` is designed around the file tree viewer in the code browser, and is why we accept a list of paths (representing open directories) and return a tree spanning the union.

An alternative, potentially simpler, approach I was going to suggest is that we could use [ls-tree](https://git-scm.com/docs/git-ls-tree) with `-r` to recurse from a specified path. It does not accept a max depth unfortunately, so we would need to truncate the output to the specified depth. Perf is ok (because of #791), even on large repos like UnrealEngine:
```
> UnrealEngine git:(release) time git ls-tree -r HEAD . > /dev/null
git ls-tree -r HEAD . > /dev/null  0.10s user 0.02s system 94% cpu 0.125 total
```

Regardless of the approach, we will want to move this api surface into the server s.t., we can also add a `list_tree` tool to the Ask agent. I'll leave that for a follow-up PR that I can open later this week.

PR #936, opened 2026-02-25
feat(web): add PostHog event for askgh repo index trigger

msukkari:
## Summary
- Adds a new `wa_askgh_repo_index_triggered` PostHog event that fires when a user triggers a new repository to be indexed in the Ask GitHub flow
- Event includes `owner`, `repo`, and `repoId` properties
- Only fires when the repo is newly created (not when visiting an already-existing repo)

## Test plan
- [ ] Visit an askgh page for a repo that hasn't been indexed yet and verify the `wa_askgh_repo_index_triggered` event appears in PostHog
- [ ] Visit an askgh page for an already-indexed repo and verify the event does NOT fire

## Summary by CodeRabbit

* **New Features**
  * Added analytics event tracking for repository indexing in Ask GitHub, capturing when users initiate repository indexing to provide insights into indexing workflow engagement.


brendan-kellam (approved):

packages/web/src/lib/posthogEvents.ts
```diff
@@ -232,6 +232,10 @@ export type PosthogEventMap = {
     },
     //////////////////////////////////////////////////////////////////
     wa_askgh_login_wall_prompted: {},
+    wa_askgh_repo_index_triggered: {
+        repo: string,
+        repoId: number,
```
brendan-kellam, line 237: nit: I would use repo name instead of the repo's id since it is opaque

PR #986, opened 2026-03-09
Add documentation workarounds for dubious ownership

joshuataylor:
This solves ownership where you're for example UID/GID 1000 but sourcebot is 1500.

This is fine because sourcebot should have readonly access.


brendan-kellam:
Heya - thanks for the PR.

I'm kind of curious as to why you are hitting dubious ownership issues when we already have a `git config --global safe.directory "*"` [in the Dockerfile](https://github.com/sourcebot-dev/sourcebot/blob/3036433e89f810d5a44827594f500a593ebfa880/Dockerfile#L240)

PR #1097, opened 2026-04-07
Layout v2

brendan-kellam:
image

image

Permission sync banner:
image

## Release Notes

* **New Features**
  * Redesigned sidebar navigation with collapsible layout replacing top navigation bar.
  * Chat history panel in sidebar with recent chats and quick actions.
  * Dedicated chats page with search, sorting, and infinite scrolling.
  * Settings reorganized with sidebar navigation for easier access.
  * General settings page for theme and editor preferences.

* **Bug Fixes & Improvements**
  * Enhanced dark mode color scheme.
  * Improved sticky search bar behavior in browse view.
  * Better tooltip positioning and animations.

* **Documentation**
  * Added guidance on conditional className composition.

PR #1106, opened 2026-04-10
v5 (#1097)

brendan-kellam:


packages/db/prisma/migrations/20260502000000_remove_guest_org_role/migration.sql
brendan-kellam, line 1: todo: re-generate this migration s.t., it comes after migrations added on main

packages/db/prisma/migrations/20260509000000_add_license_table/migration.sql
brendan-kellam, line 1: todo: re-generate this migration s.t., it comes after migrations added on main

PR #1111, opened 2026-04-13
feat(web): rename PostHog chat events, add tool_used tracking, and client-side ask events

msukkari:
## Summary

Renames server-side chat events, adds unified tool usage tracking, sets the source header on the web client chat transport, and cleans up redundant events.

### Code changes

**Events renamed** (dropped `wa_` prefix since they fire from multiple sources):
- `wa_chat_thread_created` -> `ask_thread_created` (now includes `source` property)
- `wa_chat_message_sent` -> `ask_message_sent` (now includes `source` property)
- `wa_chat_tool_used` -> `tool_used` (moved from agent.ts `onStepFinish` to the shared tool adapter layer)

**Events removed:**
- `api_code_search_request` -- redundant with `api_request` which fires for every API call and carries `path` + `source`

**New behavior:**
- `tool_used` fires from the shared adapter layer (`adapters.ts`) for every tool call, covering both ask agent (`source: 'sourcebot-ask-agent'`) and MCP server (`source: 'sourcebot-mcp-server'`) paths
- `tool_used` also fires for manually registered MCP tools (`list_language_models`, `ask_codebase`) in `server.ts`
- Web client `DefaultChatTransport` now sets `X-Sourcebot-Client-Source: sourcebot-web-client` header, so `ask_message_sent` fires with `source: 'sourcebot-web-client'` for web UI calls
- Added PostHog event naming convention to `CLAUDE.md`

### Final event inventory (for search/ask/tool workflows)

| Event | Type | Fires from | Purpose |
|-------|------|------------|---------|
| `search_finished` | client-side | `searchResultsPage.tsx` | Web-only search performance stats |
| `api_request` | server-side | `apiHandler.ts` (every API route) | Generic API traffic log with `path`, `method`, `source` |
| `ask_thread_created` | server-side | `actions.ts`, `askCodebase.ts` | Chat thread creation with `chatId`, `isAnonymous`, `source` |
| `ask_message_sent` | server-side | `chat/route.ts`, `askCodebase.ts` | Chat message with `chatId`, `messageCount`, `selectedReposCount`, `source` |
| `tool_used` | server-side | `adapters.ts`, `server.ts` | Individual tool call with `toolName`, `source`, `success` |

### PostHog dashboard configuration

**Actions (shared across all 13 company dashboards):**

**Any Feature Usage** (ID: 258475) -- used for the "All" DAU/WAU line:
1. `search_finished` (web search)
2. `wa_chat_message_sent` (legacy ask, backwards compat)
3. `ask_message_sent` where `source = 'sourcebot-web-client'` (new ask)
4. `api_request` where `source = 'mcp'` (external MCP npm package HTTP calls)
5. `tool_used` where `source = 'sourcebot-mcp-server'` (in-process MCP tool calls)

**Ask Usage** (ID: 263321) -- used for the "Ask" DAU/WAU line:
1. `wa_chat_message_sent` (legacy, backwards compat)
2. `ask_message_sent` where `source = 'sourcebot-web-client'` (new)

**MCP Usage** (ID: 263322) -- used for the "MCP" DAU/WAU line:
1. `api_request` where `source = 'mcp'` (external npm package HTTP calls)
2. `tool_used` where `source = 'sourcebot-mcp-server'` (in-process MCP tool calls)

**Insights:**

**Feature DAU** (ID: 7539703) -- 4 series: All (Any Feature Usage action), Search (`search_finished`), Ask (Ask Usage action), MCP (MCP Usage action). Daily interval, -90d range. On all 13 dashboards.

**Feature WAU** (ID: 7539704) -- Same as DAU but weekly interval.

**Tool Usage** (ID: 7919953) -- `tool_used` total count, breakdown by `toolName`. Daily interval, -90d range. On all 13 dashboards. **New.**

## Test plan

- [ ] Verify `ask_message_sent` fires with `source: 'sourcebot-web-client'` when sending a message from the web chat UI
- [ ] Verify `ask_message_sent` fires with `source: 'mcp'` when using the in-process MCP `ask_codebase` tool
- [ ] Verify `ask_thread_created` fires for both web UI and MCP/API thread creation
- [ ] Verify `tool_used` events fire with correct `toolName` and `source` for both ask agent and MCP tool calls
- [ ] Verify Feature DAU/WAU dashboards show historical data correctly (no gaps at the transition)
- [ ] Grep codebase for `wa_chat_thread_created`, `wa_chat_message_sent`, `wa_chat_tool_used`, `api_code_search_request` -- should only appear in CHANGELOG.md

### DIFF
--- a/packages/web/src/features/tools/adapters.ts
+++ b/packages/web/src/features/tools/adapters.ts
@@ -24,7 +24,7 @@ export function toVercelAITool<TName extends string, TShape extends z.ZodRawShap
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success,
-                }, { distinctId: context.userId });
+                });
             }
         },
         toModelOutput: ({ output }) => ({
@@ -67,7 +67,7 @@ export function registerMcpTool<TName extends string, TShape extends z.ZodRawSha
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success,
-                }, { distinctId: context.userId });
+                });
             }
         },
     );
--- a/packages/web/src/features/tools/adapters.ts
+++ b/packages/web/src/features/tools/adapters.ts
@@ -3,7 +3,6 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import { captureEvent } from "@/lib/posthog";
 import { ToolContext, ToolDefinition } from "./types";
-import { logger } from "./logger";
 
 export function toVercelAITool<TName extends string, TShape extends z.ZodRawShape, TMetadata>(
     def: ToolDefinition<TName, TShape, TMetadata>,
@@ -25,9 +24,7 @@ export function toVercelAITool<TName extends string, TShape extends z.ZodRawShap
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success,
-                }, { distinctId: context.userId }).catch((error) => {
-                    logger.warn(`Failed to capture tool_used event for ${def.name}:`, error);
-                });
+                }, { distinctId: context.userId });
             }
         },
         toModelOutput: ({ output }) => ({
@@ -70,9 +67,7 @@ export function registerMcpTool<TName extends string, TShape extends z.ZodRawSha
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success,
-                }, { distinctId: context.userId }).catch((error) => {
-                    logger.warn(`Failed to capture tool_used event for ${def.name}:`, error);
-                });
+                }, { distinctId: context.userId });
             }
         },
     );
--- a/packages/web/src/features/tools/adapters.ts
+++ b/packages/web/src/features/tools/adapters.ts
@@ -25,7 +25,7 @@ export function toVercelAITool<TName extends string, TShape extends z.ZodRawShap
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success,
-                }).catch((error) => {
+                }, { distinctId: context.distinctId }).catch((error) => {
                     logger.warn(`Failed to capture tool_used event for ${def.name}:`, error);
                 });
             }
@@ -63,7 +63,7 @@ export function registerMcpTool<TName extends string, TShape extends z.ZodRawSha
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success: true,
-                }).catch((error) => {
+                }, { distinctId: context.distinctId }).catch((error) => {
                     logger.warn(`Failed to capture tool_used event for ${def.name}:`, error);
                 });
                 return { content: [{ type: "text" as const, text: result.output }] };
@@ -72,7 +72,7 @@ export function registerMcpTool<TName extends string, TShape extends z.ZodRawSha
                     toolName: def.name,
                     source: context.source ?? 'unknown',
                     success: false,
-                }).catch((error) => {
+                }, { distinctId: context.distinctId }).catch((error) => {
                     logger.warn(`Failed to capture tool_used event for ${def.name}:`, error);
                 });
                 const message = error instanceof Error ? error.message : String(error);
--- a/packages/web/src/features/tools/adapters.ts
+++ b/packages/web/src/features/tools/adapters.ts
@@ -3,6 +3,7 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import { captureEvent } from "@/lib/posthog";
 import { ToolContext, ToolDefinition } from "./types";
+import { logger } from "./logger";
 
 export function toVercelAITool<TName extends string, TShape extends z.ZodRawShape, TMetadata>(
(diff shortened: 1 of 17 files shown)


packages/web/src/lib/posthogEvents.ts
```diff
@@ -286,10 +288,6 @@ export type PosthogEventMap = {
     //////////////////////////////////////////////////////////////////
     wa_repo_not_found_for_zoekt_file: {},
     //////////////////////////////////////////////////////////////////
-    api_code_search_request: {
```
brendan-kellam, line 289: could you explain why we are removing this event?

packages/web/src/features/tools/adapters.ts
```diff
@@ -41,8 +59,22 @@ export function registerMcpTool<TName extends string, TShape extends z.ZodRawSha
             try {
                 const parsed = def.inputSchema.parse(input);
                 const result = await def.execute(parsed, context);
+                captureEvent('tool_used', {
```
brendan-kellam: nit: could we do the same pattern as `toVercelAITool` where we fire the captureEvent in a `finally` block with a success boolean?

packages/web/src/lib/posthogEvents.ts
msukkari: its redundant with the api_request event thats called

packages/web/src/features/tools/adapters.ts
```diff
@@ -11,7 +13,23 @@ export function toVercelAITool<TName extends string, TShape extends z.ZodRawShap
...
+                return await def.execute(input, context);
+            } catch (error) {
+                success = false;
+                throw error;
+            } finally {
+                captureEvent('tool_used', {
+                    toolName: def.name,
+                    source: context.source ?? 'unknown',
+                    success,
+                }).catch((error) => {
```
brendan-kellam: do we need a .catch on captureEvent? Maybe we want to move the exception handling into `captureEvent` itself s.t., we are guaranteed that it won't throw when calling it? (which imo is a reasonable assumption since capturing events should fail gracefully)

packages/web/src/lib/posthogEvents.ts
```diff
@@ -159,24 +159,26 @@ export type PosthogEventMap = {
         chatId: string,
         messageId: string,
     },
-    wa_chat_thread_created: {
+    ask_thread_created: {
```
brendan-kellam, line 162: In posthog, is there a way to map the old event name (`wa_chat_thread_created`) to the new event name (`ask_thread_created`)? My only concern with renaming events is that insight we create with the new event name will not capture deployments that are still using the old event name.

msukkari: Yes, we use actions in Posthog which combine the old and new event

packages/web/src/features/tools/adapters.ts
msukkari: done

packages/web/src/features/tools/adapters.ts
msukkari: moved try/catch to inside captureEvent

packages/web/src/lib/posthog.ts
```diff
@@ -87,25 +87,29 @@ export const createPostHogClient = async () => {
...
+            properties: {
+                ...properties,
+                sourcebot_version: SOURCEBOT_VERSION,
+                install_id: env.SOURCEBOT_INSTALL_ID,
+                $host: host,
+            },
+            distinctId,
+        });
+    } catch {
+        // Telemetry should never break application functionality.
```
brendan-kellam: can we add a error log?

packages/web/src/lib/posthog.ts
```diff
@@ -87,25 +87,29 @@ export const createPostHogClient = async () => {
...
 }
 
-export async function captureEvent<E extends PosthogEvent>(event: E, properties: PosthogEventMap[E]) {
-    if (env.SOURCEBOT_TELEMETRY_DISABLED === 'true') {
-        return;
-    }
-
-    const distinctId = await tryGetPostHogDistinctId();
-    const posthog = await createPostHogClient();
+export async function captureEvent<E extends PosthogEvent>(event: E, properties: PosthogEventMap[E], options?: { distinctId?: string }) {
```
brendan-kellam: why do we need to pass in the distinctId here?

brendan-kellam: My assumption would be that `tryGetPostHogDistinctId` would be able to determine the distinctId from 1) the posthog cookie for ask, and 2) the session object for mcp

msukkari: it previously didn't but I just made it so that it does by using the existing `getAuthenticatedUser` method - removed all of this user id piping

packages/web/src/lib/posthog.ts
msukkari: added

brendan-kellam (approved):

packages/web/src/lib/posthog.ts
brendan-kellam: lgtm

Issue #1016, opened 2026-03-19
[FR]Sort tags and branches list before 64-revision cap so pruning keeps newest N refs


GitBalake:
When tags and branches match many refs, Sourcebot truncates to 64 revisions using `revisions.slice(0, 64)` after listing tags via plain `git.tags()` / `git tag` ordering. That order is ref-name order, not recency. For many repos this is naturally oldest tags first.  I am guessing this is the opposite of what many users want.

Git already supports explicit ordering (e.g. `git tag --sort=-creatordate`, `--sort=-committerdate`, `--sort=-version:refname`). This request is to apply an explicit sort before truncation. And ideally have that exposed in the config. 

## Relation to unlimited branches / tags

There is already ongoing work toward removing the 64 revision cap [#461](https://github.com/sourcebot-dev/sourcebot/issues/461) and related PRs). That helps orgs that want a full ref index.

Even with unlimited or much higher limits, a lot of people will still want optionally capped index size and cost. A newest N tags/branches or newest by last commit date policy seems like the natural knob to pass in. 

Might be worth adding that sort option and keeping a cap in that other branch.

Currently right now I get my oldest x tags while working on software that matches live code to its sourcebot index for that tag. Naturally I am missing all tags.

brendan-kellam:
thanks for the FR, this is good feedback - I agree that adding a `sort` option here that allows you to specify one of creation date, commit date, or reference name makes sense. I will try to get to this soon, also open to a PR here as well if you have time, just lmk

GitBalake:
I did not see your response. If this has not been referenced in any other work I am open to a PR. I also know that there was a similar open issue for endless tags. So I worry they will conflict. I also worry about indexes of unlimited tags without configurable limits. 

@brendan-kellam let me know and I will get back to you faster this time.

brendan-kellam:
@GitBalake - we are not actively working on this, so if you have time, a PR would be welcomed. It would be helpful if you could first outline what config changes you would make s.t., we can align on what is required.

> I also worry about indexes of unlimited tags without configurable limits.

If you are referring to #461, the current plan is that, by default, we will **only** create a index for the default branch. All other branches, tags, commit SHAs, etc. will be powered by plaintext search (i.e., `git grep`).

[...]

PR #1122, opened 2026-04-15
fix(worker): prefer newest refs when applying the 64 revision cap

GitBalake:
Part of #1016

## Summary

Sourcebot currently caps explicit branch/tag indexing at 64 revisions. Before this change, matching refs were effectively pruned in ref-name order, which could cause older tags to be kept while newer tags were dropped.

This change applies sorting before truncation so the cap keeps the newest matching refs by default:

- branches are ordered by committer date
- tags are ordered by creator date
- the default branch remains pinned first

## Verification

- Added backend tests covering newest-first branch and tag selection
- Added regression coverage for default-branch pinning and de-duplication before truncation
- Passed all backend testing
- Ran backend build
- Verified live indexing on a real repo with more than 64 matching refs and confirmed Sourcebot stored newest-first indexed revisions

## Summary by CodeRabbit

* **Bug Fixes**
  * Revision selection now prefers the newest matching branches and tags, de-duplicates entries, and consistently limits the list to 64 revisions.

* **Tests**
  * Added tests verifying branch and tag ordering by timestamp and confirming newest-first revision selection and truncation behavior.

* **Documentation**
  * Updated changelog to document the revision-selection fix.


brendan-kellam (approved):
LGTM thanks! Appreciate the addition of git.test.ts - we could use more coverage on our git operations, so this is a great first step

PR #1130, opened 2026-04-17
fix(backend): stop Gitea/Forgejo pagination on empty page response

### LINKED ISSUE #1108: [bug] Infinite API pagination loop when Gitea/Forgejo token cannot see all org repos
### Describe the bug

When using a Gitea/Forgejo connection with "orgs" config, if the API token's user does not have access to all
  repositories in the organization, Sourcebot enters an infinite pagination loop. It calls GET
  /api/v1/orgs/{org}/repos?limit=100&page=N with N incrementing indefinitely (observed page numbers exceeding 330,000)
  at 10+ requests per second, effectively DoS-ing the Forgejo instance.

  The Forgejo API reports the org's total repo count in its response, but the token can only see a subset. Sourcebot
  appears to keep paginating because the number of collected repos never matches the reported total.

  Expected behavior: Sourcebot should stop paginating when the API returns an empty page, regardless of the reported
  total count.

### To reproduce

1. Set up a Gitea/Forgejo org with multiple repos (e.g. 13)
  2. Create a user and API token that only has collaborator access to some repos in the org (not org-level member)
  3. Configure Sourcebot with "orgs" pointing to that org:
  {
      "connections": {
          "forgejo": {
              "type": "gitea",
              "url": "https://forgejo.example.com",
              "token": { "env": "GITEA_TOKEN" },
              "orgs": ["MyOrg"]
          }
      }
  }
  4. Start Sourcebot
  5. Observe the Gitea/Forgejo access logs — page numbers will climb into the hundreds of thousands and never stop

Sourcebot v4.16.8 (also reproduced on v4.15.11). Self-hosted via Docker Compose against a self-hosted Forgejo (12.0.1+gitea-1.22.0)
  instance.

### Additional information

Impact is severe — the constant API requests (each taking 600-2000ms) cause Forgejo to become unresponsive with slow
  SQL queries (5-9 seconds), 500 errors on the web UI, and degraded service for all users and CI systems. Sourcebot's
  own web UI also starts throwing NOT_AUTHENTICATED errors due to the backend being overwhelmed.

  Workaround: Give the token's user org-level read access to all repositories (via an org team with "All repositories"
  permission) so the visible repo count matches the reported total.

brendan-kellam:
When an API token can only see a subset of org repos, the `x-total-count` header reports the org total while the token returns fewer items. The previous loop condition (`output.length < totalCount`) would never be satisfied for the visible subset, causing infinite pagination and effectively DoS-ing the Gitea/Forgejo instance.

The fix breaks out of the loop when the API returns an empty page, matching the Gitea API pagination docs recommendation.

Fixes #1108

Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed infinite pagination loops in Gitea/Forgejo integrations that occurred when API tokens had restricted access to certain repositories. The pagination system now correctly detects when all available results have been retrieved and terminates appropriately, preventing endless looping and excessive API requests while significantly improving overall system stability, reliability, and performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed revision selection so the 64-revision cap prefers the newest matching branches and tags instead of pruning by ref-name order. [#1122](https://github.com/sourcebot-dev/sourcebot/pull/1122)
+- Fixed infinite pagination loop in Gitea/Forgejo when an API token can only see a subset of org repos (the `x-total-count` header reports org total while token returns fewer items). [#1130](https://github.com/sourcebot-dev/sourcebot/pull/1130)
 
 ## [4.16.11] - 2026-04-17
 
--- a/packages/backend/src/gitea.ts
+++ b/packages/backend/src/gitea.ts
@@ -254,6 +254,9 @@ const paginate = async <T>(request: (page: number) => Promise<HttpResponse<T[], 
     while (output.length < totalCount) {
         page++;
         const result = await request(page);
+        if (result.data.length === 0) {
+            break;
+        }
         output.push(...result.data);
     }
 


brendan-kellam:
@claude add a changelog entry

PR #1138, opened 2026-04-20
fix(query-language): preserve grouped filters in regex mode

brianphillips:
## Summary
- preserve query grouping in regex mode when parentheses wrap query syntax instead of raw regex atoms
- keep bare regex groups like `(foo|bar)` parsed as single regex terms
- add regression coverage for grouped prefix filters and filtered parenthesized regex `or` groups

## Testing
- `../../node_modules/.bin/vitest --run` (from `packages/queryLanguage`)

## :robot: Agent Plan
<details>
<summary>Expand full agent plan</summary>

1. Trace how `isRegexEnabled` changes query parsing and identify where field-specific filters are converted into regex queries.
2. Reproduce the failing parse shape and confirm regex mode is treating grouped filter parentheses as part of a regex term instead of query grouping.
3. Update the query-language tokenizer so regex mode distinguishes raw regex parentheses from query-like parenthesized groups, including negated groups.
4. Add regression tests for grouped prefix filters, grouped regex OR expressions behind filters, and the nested filter plus `(regex or regex)` grouping discussed in this thread.
5. Run the focused `packages/queryLanguage` Vitest suite and verify the failing queries parse correctly.
6. Exclude unrelated repository state changes, including the `vendor/zoekt` submodule rewind, from the PR.

</details>

## Release Notes

* **Bug Fixes**
  * Corrected regex mode search parsing to properly handle parenthesized filter groups. Query-style grouping syntax now works correctly when regex mode is enabled, allowing users to combine multiple filtered conditions with regex patterns for more precise and complex searches.


brendan-kellam (approved):
LGTM thx!

PR #1145, opened 2026-04-23
fix(web): restore ServiceError boundary in `getFileSourceForRepo`

### LINKED ISSUE #1144: [bug] v4.16.14 results in `EACCESS permission denied` errors
### Describe the bug

Upgraded to `v4.16.14` and encountered the following error:

```
[web] | 2026-04-23T16:09:55.758Z error: [sew] Error: EACCES: permission denied, stat '/data/.sourcebot/repos/131' {"task":{"commands":[],"format":"empty"}}
[web] | Error: Error: EACCES: permission denied, stat '/data/.sourcebot/repos/131'
[web] |     at z.onFatalException (/app/packages/web/.next/server/chunks/_08l9pj1._.js:2:15644)
[web] |     at z.attemptTask (/app/packages/web/.next/server/chunks/_08l9pj1._.js:2:15536)
```

Downgraded to `v4.16.13` and errors went away... 

Looks like my changes in #1104 caused an unintended breakage. 

Confirmed by Sourcebot - https://app.sourcebot.dev/chat/cmobokddf0075k36xu33uu022

### To reproduce

Upgrade to `v4.16.14`

Sourcebot version (e.g. v3.0.1): `v4.16.14`

### Additional information

Fix PR incoming.

fatmcgav:
## Problem

In v4.16.14 (PR #1104, GitLab MR Review Agent), the core file-fetch logic was
extracted from inside `sew()` into a standalone `getFileSourceForRepo` function
so it could be called by privileged server-side callers (e.g. the review agent
webhook handler) without going through the auth middleware.

The extraction introduced a missing error boundary. The catch block inside
`getFileSourceForRepo` handled two known git error patterns and **re-threw
everything else**:

```ts
// getFileSourceApi.ts — before this fix
} catch (error: unknown) {
    const errorMessage = error instanceof Error ? error.message : String(error);
    if (errorMessage.includes('does not exist') || ...) return fileNotFound(...);
    if (errorMessage.includes('unknown revision') || ...) return unexpectedError(...);
    throw error;   // ← propagates uncaught; sew() no longer wraps this code path
}
```

Because `getFileSourceForRepo` is no longer inside `sew()`, any re-thrown
exception escapes as a fatal error through the Next.js server task runner,
producing the `z.onFatalException` / `z.attemptTask` stack trace seen in
production on v4.16.14.

A second contributing factor: the review agent was also changed in v4.16.14 to
pass `ref: pr_payload.head_sha` where `ref` was previously `undefined`. If the
bare clone hasn't fetched that commit yet, `git show` throws with
`"unknown revision"` — which fell into the same unhandled re-throw path.

Rolling back to v4.16.13 restored the original `sew()`-wrapped code path, which
is why the rollback fixed the issue.

 ## Fix

Two targeted changes in `getFileSourceApi.ts`:

1. **`throw error` → `return unexpectedError(errorMessage)`** — ensures
   `getFileSourceForRepo` always returns `FileSourceResponse | ServiceError`
   and never rejects its promise. This is the root-cause fix.

2. **`unexpectedError(...)` → `invalidGitRef(gitRef)`** for the
   `"unknown revision"` / `"bad revision"` / `"invalid object name"` branch —
   uses the semantically correct error type (already imported), which also
   gives callers a more actionable error when `head_sha` hasn't been fetched.

 ## Tests

Added `getFileSourceApi.test.ts` with 19 tests covering:

- **Repository validation** — `NOT_FOUND` when repo is absent from the DB;
  correct `findFirst` query shape.
- **Input validation** — path traversal and null-byte paths → `FILE_NOT_FOUND`;
  refs starting with `-` → `INVALID_GIT_REF` (flag-injection guard).
- **Git error handling** (the regression suite):
  - `"does not exist"` / `"fatal: path"` → `FILE_NOT_FOUND`
  - `"unknown revision"` (unfetched `head_sha`) → `INVALID_GIT_REF`
  - `"bad revision"` / `"invalid object name"` → `INVALID_GIT_REF`
  - **Unrecognised error → `UNEXPECTED_ERROR`, not a throw** — explicit
    regression test; before the fix, `.resolves` would fail because the
    promise rejected.
- **Successful response** — source content, language, ref fallback chain
  (`ref` → `defaultBranch` → `HEAD`), correct `cwd` path.
- **Language detection** — prefers `.gitattributes`; falls back to
  filename-based detection when the file is absent.

 ## Files changed

| File | Change |
|------|--------|
| `packages/web/src/features/git/getFileSourceApi.ts` | `throw error` → `return unexpectedError(errorMessage)`; invalid-ref branch now returns `invalidGitRef(gitRef)` |
| `packages/web/src/features/git/getFileSourceApi.test.ts` | New — 19 tests |

Fixes #1144

## Summary by CodeRabbit

* **Bug Fixes**
  * All error paths now return consistent, user-facing service errors (no uncaught exceptions). Unresolved git references produce a clear, specific error and git failures are mapped to appropriate, user-facing responses; function now guarantees a resolved error result.

* **Tests**
  * Added comprehensive tests for file-retrieval flows: repo lookup, safety checks, git error mappings, ref handling (ref → defaultBranch → HEAD), and language/source resolution.

### DIFF
--- a/packages/web/src/features/git/getFileSourceApi.ts
+++ b/packages/web/src/features/git/getFileSourceApi.ts
@@ -26,84 +26,79 @@ export type FileSourceResponse = z.infer<typeof fileSourceResponseSchema>;
 export const getFileSourceForRepo = async (
     { path: filePath, repo: repoName, ref }: FileSourceRequest,
     { org, prisma }: { org: Org; prisma: PrismaClient },
-): Promise<FileSourceResponse | ServiceError> => {
+): Promise<FileSourceResponse | ServiceError> => sew(async () => {
+    const repo = await prisma.repo.findFirst({
+        where: { name: repoName, orgId: org.id },
+    });
+    if (!repo) {
+        return notFound(`Repository "${repoName}" not found.`);
+    }
+
+    if (!isPathValid(filePath)) {
+        return fileNotFound(filePath, repoName);
+    }
+
+    if (ref !== undefined && !isGitRefValid(ref)) {
+        return invalidGitRef(ref);
+    }
+
+    const { path: repoPath } = getRepoPath(repo);
+    const git = simpleGit().cwd(repoPath);
+
+    const gitRef = ref ?? repo.defaultBranch ?? 'HEAD';
+
+    let fileContent: string;
     try {
-        const repo = await prisma.repo.findFirst({
-            where: { name: repoName, orgId: org.id },
-        });
-        if (!repo) {
-            return notFound(`Repository "${repoName}" not found.`);
-        }
-
-        if (!isPathValid(filePath)) {
-            return fileNotFound(filePath, repoName);
-        }
-
-        if (ref !== undefined && !isGitRefValid(ref)) {
-            return invalidGitRef(ref);
-        }
-
-        const { path: repoPath } = getRepoPath(repo);
-        const git = simpleGit().cwd(repoPath);
-
-        const gitRef = ref ?? repo.defaultBranch ?? 'HEAD';
-
-        let fileContent: string;
-        try {
-            fileContent = await git.raw(['show', `${gitRef}:${filePath}`]);
-        } catch (error: unknown) {
-            const errorMessage = error instanceof Error ? error.message : String(error);
-            if (errorMessage.includes('does not exist') || errorMessage.includes('fatal: path')) {
-                return fileNotFound(filePath, repoName);
-            }
-            if (errorMessage.includes('unknown revision') || errorMessage.includes('bad revision') || errorMessage.includes('invalid object name')) {
-                return unresolvedGitRef(gitRef);
-            }
-            return unexpectedError(errorMessage);
-        }
-
-        let gitattributesContent: string | undefined;
-        try {
-            gitattributesContent = await git.raw(['show', `${gitRef}:.gitattributes`]);
-        } catch {
-            // No .gitattributes in this repo/ref, that's fine
-        }
-
-        const language = gitattributesContent
-            ? (resolveLanguageFromGitAttributes(filePath, parseGitAttributes(gitattributesContent)) ?? detectLanguageFromFilename(filePath))
-            : detectLanguageFromFilename(filePath);
-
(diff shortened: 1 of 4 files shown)


packages/web/src/features/git/getFileSourceApi.ts
```diff
@@ -27,77 +27,82 @@ export const getFileSourceForRepo = async (
     { path: filePath, repo: repoName, ref }: FileSourceRequest,
     { org, prisma }: { org: Org; prisma: PrismaClient },
 ): Promise<FileSourceResponse | ServiceError> => {
-    const repo = await prisma.repo.findFirst({
-        where: { name: repoName, orgId: org.id },
-    });
-    if (!repo) {
-        return notFound(`Repository "${repoName}" not found.`);
-    }
+    try {
```
brendan-kellam: why are we wrapping this in a try block?

brendan-kellam: could we just use a `sew` on this function as well?

fatmcgav: This came out of a Coderabbit suggestion - https://github.com/sourcebot-dev/sourcebot/pull/1145#pullrequestreview-4164433846

fatmcgav: With some more prompting, it looks like `sew()` might be a better wrapper here though :)

fatmcgav: > could we just use a `sew` on this function as well?

Missed this comment.. But yeh, that was the end result 👍

brendan-kellam (approved):
LGTM thanks!

The review agent is a unique path where we don't actually have a auth context to lean on (and therefore, cannot use withAuth).

Some loose thoughts: Eventually, I would like to get to the point where there is some amount of formalization between the "API layer" where calls are authenticated and authorized for the caller, and the "service layer" where the business logic lives. The idea would then be that we could better reason about what code does / does not contain auth guarantees.

For the most part, this formalization is not necessary right now, but it's something to consider

PR #1149, opened 2026-04-24
fix(web): fix GitLab MR inline comment 400 errors on context lines and renames

fatmcgav:
Two bugs caused `MergeRequestDiscussions.create` to return 400 Bad Request for a subset of inline comments.

---

### Missing `oldPath` on modified files

GitLab's Discussions API requires both `oldPath` and `newPath` in the position object for any file that exists in both the base and head commits. Previously only `newPath` was set, causing GitLab to reject comments on modified files. `oldPath` is now always included, falling back to `filename` when no rename is present.

### Missing `oldLine` for context (unchanged) lines

GitLab requires `oldLine` in addition to `newLine` when the target line is a context line (unchanged in the diff). Without it the API returns 400 for those lines while accepting added-line comments — explaining why only *some* comments in a given MR failed.

The fix parses `oldSnippet`/`newSnippet` from the diff payload to build a new→old line number map. Context lines appear at the same index in both snippets, so zipping them gives the mapping. `oldLine` is only included when the line is confirmed to be a context line; added lines continue to use `newLine` only.

### Rename support

`oldFilename` is now propagated through `generatePrReview.ts` → `sourcebot_file_diff_review` → `gitlabPushMrReviews`, so renamed files correctly pass the old path as `oldPath`.

---

### Files changed

| File | Change |
|------|--------|
| `types.ts` | Added `oldFilename?: string` to `sourcebot_file_diff_review_schema` |
| `generatePrReview.ts` | Sets `oldFilename: file_diff.from` when building review objects |
| `gitlabPushMrReviews.ts` | Adds `oldPath`, context-line `oldLine` mapping via `buildContextLineMap` |
| `gitlabPushMrReviews.test.ts` | Adds `oldPath` to position assertion; new tests for renames and context vs added line behaviour |

Related: #1104

## Summary by CodeRabbit

* **New Features**
  * Reviews now record original file paths and map new-line coordinates to their corresponding old-line coordinates so comments retain context across renamed files.

* **Bug Fixes**
  * Prevent GitLab inline review failures when commenting on context/unchanged lines or renamed files by supplying old-path and, where applicable, old-line information in comment positions.


fatmcgav:
@brendan-kellam @msukkari Hopefully this one is a quick win to merge :)

brendan-kellam (approved):
lgtm thanks!

brendan-kellam (approved):

PR #1219, opened 2026-05-22
fix(web): propagate session invalidation to /api/auth/session

msukkari:
- bump `sessionVersion` when user signs out
- add `sessionVersion` check in `jwt` callback
- remove redundant sessionVersion check in `auth` method, which now uses `sessionVersion` check in the `jwt` callback

Ensures that:
- logging out invalidates all previous session tokens (even if they were captured outside the browser)
- ensures that calls to `/api/auth/session` are covered by session version check

## Bug Fixes

* Session invalidation events—including logout, user deletion, and removal from an organization—are now properly reflected in real-time session status checks.
* Previously issued authentication tokens are now correctly invalidated upon logout, preventing their reuse in subsequent API requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sourcebot-dev/sourcebot/pull/1219?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/web/src/auth.ts
+++ b/packages/web/src/auth.ts
@@ -214,6 +214,15 @@ const nextAuthResult = NextAuth({
         signOut: async (message) => {
             const token = message as { token: { userId: string } | null };
             if (token?.token?.userId) {
+                // Bump sessionVersion so any JWT minted before this signout
+                // is treated as invalid by the jwt callback's DB cross-check
+                // on its next request, even if the cookie value was captured
+                // and is being replayed.
+                await __unsafePrisma.user.update({
+                    where: { id: token.token.userId },
+                    data: { sessionVersion: { increment: 1 } },
+                });
+
                 await auditService.createAudit({
                     action: "user.signed_out",
                     actor: {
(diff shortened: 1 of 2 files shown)


packages/web/src/auth.ts
```diff
@@ -313,35 +352,18 @@ const nextAuthResult = NextAuth({
...
- * session as logged out.
+ * JWT validity (including the `sessionVersion` cross-check against the
+ * database and the existence of the underlying `User` row) is enforced in
+ * the `jwt` callback above. If that callback returns `null`, NextAuth's
+ * core resolves the session to `null` here and also clears the cookie on
+ * the response. We therefore only need to memoize the result within a
+ * single request so that multiple `auth()` callers share the same answer
+ * without re-running the upstream resolver.
  */
 export const auth = cache(async (): Promise<Session | null> => {
```
brendan-kellam, line 366: seems like we can remove this now?

packages/web/src/auth.ts
```diff
@@ -259,20 +268,50 @@ const nextAuthResult = NextAuth({
...
+                }
+
+                // On every non-login request, cross-check the JWT's
+                // sessionVersion against the user's current sessionVersion in
+                // the database. A mismatch means the user signed out, was
+                // removed from the org, or their sessions were otherwise
+                // invalidated since the JWT was minted. Returning null here
+                // is what makes invalidation visible at /api/auth/session,
+                // not just at withAuth-gated endpoints.
+                if (!user && (token.sessionVersion ?? 0) !== dbUser.sessionVersion) {
```
brendan-kellam: what was the motivation in moving this check from the wrapped session resolver into jwt?

brendan-kellam: nit: could we move `token.sessionVersion ?? 0` out of this conditional into a seperate variable? Feel like this is slightly hard to read and has the potential to be misinterpreted

packages/web/src/auth.ts
brendan-kellam: https://react.dev/reference/react/cache

packages/web/src/auth.ts
msukkari: so that the `api/auth/session` endpoint also goes through the sessionVersion check, since that endpoint doesn't go through `auth`

brendan-kellam (approved):

PR #1246, opened 2026-05-29
fix: reconcile missing zoekt shards on startup - Fixes #1210

### LINKED ISSUE #1210: [bug/rfe] Rebuild or mark repos stale when zoekt shard files are missing but DB marks repos indexed
## Summary

If the zoekt index directory is deleted while the Sourcebot database and repo clones remain intact, Sourcebot starts successfully but does not automatically rebuild the missing shards. Repos remain marked as indexed in the DB, while zoekt has no shard files to search, so search returns incomplete or empty results from an otherwise healthy instance.

This matters for Kubernetes deployments that want to place only `.sourcebot/index` on more performant ephemeral/local storage while keeping DB/repos on durable storage.

## Version

Observed on:

- Sourcebot image: `ghcr.io/sourcebot-dev/sourcebot:v4.17.1`

## Background / Use Case

For performance reasons, we are evaluating a hybrid storage layout:

- durable PVC for `/data`
- node-local / ephemeral storage for `/data/.sourcebot/index`

The zoekt index is derived/rebuildable, while the DB and repo clones should remain durable. This can improve cold/random read performance for large zoekt indexes, but pod replacement would delete only the index directory.

In that scenario, Sourcebot should ideally detect that indexed repos no longer have shard files and enqueue rebuilds, or otherwise mark those repos as not indexed.

## Current Behavior

When `/data/.sourcebot/index` is empty but the DB still has repos with fresh `indexedAt` values:

- Sourcebot starts normally.
- zoekt tolerates the empty index directory.
- The repo remains marked as indexed/completed in the DB.
- No new indexing job is scheduled.
- Search returns zero/incomplete results.
- Manual reindex through `/api/index-repo` successfully rebuilds the missing shard.

## Expected Behavior

On startup, or during repo-index scheduling, Sourcebot should detect repos that are marked indexed in the DB but have no corresponding zoekt shard files.

Possible expected behaviors:

- enqueue index jobs for affected repos,
- set `indexedAt = null` for affected repos so the existing scheduler picks them up,
- or expose a supported startup/recovery command that rebuilds all missing index shards.

## Relevant Code Paths

From a quick read of the code:

- `packages/backend/src/constants.ts`
  - `REPOS_CACHE_DIR = DATA_CACHE_DIR/repos`
  - `INDEX_CACHE_DIR = DATA_CACHE_DIR/index`

- `packages/backend/src/index.ts`
  - creates the repos and index directories at startup if absent

- `packages/backend/src/repoIndexManager.ts`
  - `scheduleIndexJobs()` selects repos where `indexedAt` is `null` or older than `reindexIntervalMs`
  - `cleanupOrphanedDiskResources()` removes shard files with no corresponding DB repo, but does not detect the inverse: DB repo exists and is marked indexed, but shard files are missing

- `packages/backend/src/zoekt.ts`
  - `indexGitRepository()` invokes `zoekt-git-index -index ${INDEX_CACHE_DIR}`

## Suggested Fix

Add startup or scheduler reconciliation that compares DB repo state with shard files in `INDEX_CACHE_DIR`.

For each repo with `indexedAt != null`, verify that at least one shard exists with that repo's shard prefix. If no shard exists, either:

1. enqueue an index job for that repo, or
2. mark the repo as stale by clearing `indexedAt`, allowing the existing scheduler to enqueue it.

This would make derived/ephemeral index storage safe to use with durable DB/repos.

## Reproduction

Use a small Sourcebot instance with a local `file://` repo.

Example config:

```json
{
  "connections": {
    "local-repos": {
      "type": "git",
      "url": "file:///data/local-repos/*"
    }
  },
  "settings": {
    "reindexIntervalMs": 86400000,
    "reindexRepoPollingIntervalMs": 1000,
    "maxRepoIndexingJobConcurrency": 1
  }
}
```

The long `reindexIntervalMs` is intentional so we can distinguish immediate missing-shard recovery from the normal periodic reindex.

Steps:

1. Create a local git repo under `/data/local-repos/testrepo`.
2. Start Sourcebot.
3. Wait for the repo to index successfully.
4. Confirm:
   - a shard exists under `/data/.sourcebot/index`
   - the repo has `indexedAt` set
   - search returns a result from the repo
5. Delete only the zoekt shard files:

```bash
rm -f /data/.sourcebot/index/*.zoekt \
      /data/.sourcebot/index/*.zoekt.meta \
      /data/.sourcebot/index/*.tmp
```

6. Restart Sourcebot with the same DB and repo data.
7. Observe that no new index job is created and search returns no results.

Manual recovery works:

```bash
curl -s -X POST http://localhost:3060/api/index-repo \
  -H "Content-Type: application/json" \
  -d '{"repoId":1}'
```

After this, the shard is rebuilt and search works again.

`ghcr.io/sourcebot-dev/sourcebot:v4.17.1`

### Additional information

_No response_

Shashank200345:
1. What was the problem

When the zoekt index directory gets wiped — say during a pod restart 
with ephemeral storage — the database still holds onto the old 
indexedAt timestamps. The scheduler sees those timestamps and assumes 
everything is fine, so it never triggers a re-index. Search returns 
nothing but the UI still shows the repo as Completed. No error, no 
warning, just silently broken.

2. Why it happened

cleanupOrphanedDiskResources() already handles one side of this — 
it cleans up shard files on disk that have no matching DB record. 
But nobody was checking the other direction: a repo marked as indexed 
in the DB with no actual shard file on disk.

3. What I changed

Added reconcileMissingShards() which runs at startup right after 
cleanupOrphanedDiskResources(). It reads what shard files actually 
exist on disk, cross checks against repos the DB thinks are indexed, 
and clears indexedAt for anything that is missing a shard. The 
existing scheduler then picks those repos up on its next poll and 
re-indexes them automatically.

Only one file changed: packages/backend/src/repoIndexManager.ts

Reuses the existing getRepoIdFromShardFileName() utility so the 
shard parsing logic stays consistent with the rest of the file.

4. Proof

Logs on restart — fix detects the missing shard and resets the repo:

Screenshot 2026-05-30 002144

DB right after restart — indexedAt cleared to null:

Screenshot 2026-05-30 005447

DB after the scheduler runs — fresh timestamp, repo re-indexed automatically:

Screenshot 2026-05-30 003659

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where repositories marked as indexed could be skipped when on-disk index data was missing. The scheduler now detects missing index files, resets the indexed state for affected repositories, and ensures they are re-indexed so search and retrieval remain accurate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sourcebot-dev/sourcebot/pull/1246?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Shashank200345:
Hi, I've addressed the CodeRabbit feedback by adding a TTL check 
using reindexIntervalMs to avoid resetting repos that legitimately 
produce zero shards. Happy to make any other changes if needed!

musabshak:
@Shashank200345 Out of curiosity - what prompted you to post this PR? Had you read through the issue I created a couple of weeks ago [1] or did you independently discover / come across this "bug"?

[1] https://github.com/sourcebot-dev/sourcebot/issues/1210

Shashank200345:
> @Shashank200345 Out of curiosity - what prompted you to post this PR? Had you read through the issue I created a couple of weeks ago [1] or did you independently discover / come across this "bug"?
> 
> [1] #1210

okay so i was just going through the issues and saw this bug , my primary goal was to contribute only to "bugs" labelled issues, and then i just read this issue summary and the way you summarized the issue was clear and crisp and then i just reproduced the bug in my local system and asked claude for the possible approaches i can follow to fix this .

PR #1266, opened 2026-06-03
Add support for MCP scopes & tool permissions

jsourcebot:
- We now support the ability for an admin the set the scopes of a connector which allows proper support for servers like slack or GitHub. 
- Scopes can be modified by the admin which will invalidate all existing tokens
- Admin can now configure per tool permissions for a given connector
- Per connector/user tools are cached on redis with 1h TTL. This means a user engaging in a chat does not have to live refetch all the tools for all the connectors they have access to on every question they ask.


packages/db/prisma/migrations/20260603030027_mcp_server_tools_and_scopes/migration.sql
```diff
@@ -0,0 +1,43 @@
+/*
```
brendan-kellam: do we want to collapse this into the existing migration file?

packages/web/src/ee/features/chat/mcp/actions.ts
```diff
@@ -244,49 +305,147 @@ export const createStaticOAuthMcpServer = async (
...
-                entryPoint: 'workspace_settings',
-                serverId: mcpServer.id,
-                serverUrl: mcpServer.serverUrl,
-                authMode: getMcpAuthMode(McpServerClientInfoSource.DYNAMIC),
-            });
+export const updateMcpServerToolPermissions = async (
+    serverId: string,
+    tools: { toolName: string; permission: McpServerToolPermission }[],
+) => {
+    const parsed = updateMcpServerToolPermissionsSchema.safeParse({ serverId, tools });
```
brendan-kellam, line 333: why do we need to parse the inputs here? Presumably since this is a server action, they are already well typed?

packages/web/src/ee/features/chat/mcp/mcpToolSets.ts
```diff
@@ -93,6 +102,76 @@ function getMcpToolFailureReason(error: unknown): string {
...
+        client.serverId,
+        getScopeHash(client.requestedScopes),
+        client.serverUpdatedAt.getTime(),
+    ].join(':');
+}
+
+async function getCachedListTools(cacheKey: string): Promise<ListToolsResult | undefined> {
+    try {
+        const cached = await redis.get(cacheKey);
+        return cached ? JSON.parse(cached) as ListToolsResult : undefined;
```
brendan-kellam, line 133: could there be a case where the ListToolsResult type can change? Do we need to do some schema validation here?

packages/db/prisma/schema.prisma
```diff
@@ -693,12 +700,26 @@ model McpServer {
   @@unique([orgId, sanitizedName])
 }
 
-/// Lifetime tool call counters for an MCP server.
-model McpServerToolCallCount {
+/// OAuth scope configuration for an MCP server.
+model McpServerScope {
```
brendan-kellam: rename to McpServerOauthScope

PR #1271, opened 2026-06-03
fix(setup-wizard): keep users out of GitHub orgs / GitLab groups

msukkari:
## Problem

In the setup wizard, the GitHub **organization** search (and the GitLab **group** search) could end up with a *user* account in the final config. The org/group autocomplete offers the literal typed value when there are no API matches, and there was no check that the selection was actually an org/group. A user account sent as an org/group causes indexing to fail.

## Fix

Each selected entry is now validated on submit against the authoritative account-type API, and a confirmed wrong-type entry is rejected with a message pointing the user to the correct prompt:

- **GitHub** — `GET /users/{login}` returns `type: "User" | "Organization"`. The org prompt rejects confirmed users; the users prompt rejects confirmed orgs.
- **GitLab** — `GET /users?username=` confirms a user (exact match) and `GET /groups/{path}` confirms a group (nested `a/b` paths are always subgroups). The groups prompt rejects confirmed users; the users prompt rejects confirmed groups.

Entries that can't be resolved (offline, rate-limited, or a private namespace the token can't see) resolve to `unknown` and are allowed through, so the self-hosted / no-token path keeps working. Results are cached per login/path.

Gitea, Bitbucket, Azure DevOps, and Gerrit use free-text input (no API-backed search), so they don't have this leak and are unchanged.

## Testing

Validated the classifier helpers against the live public APIs:

- GitHub: `docker` → Organization, `torvalds` → User, nonexistent → unknown.
- GitLab: `gitlab-org` → group, `jramsay` → user, `gitlab-org/api` (subgroup) → group, nonexistent → unknown.

`tsc --noEmit` and `yarn build` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Checkout the [migration guide](https://docs.sourcebot.dev/docs/upgrade/v4-to-v5-
 
 ### Fixed
 - Fixed git "dubious ownership" errors when the container runs as a non-root user by setting `safe.directory` at the system level instead of the global (root-only) level. [#1106](https://github.com/sourcebot-dev/sourcebot/pull/1106)
+- Fixed the `setup-sourcebot` wizard accepting a user account in the GitHub organizations (or GitLab groups) list by validating each selection against the code host's account-type API. [#1271](https://github.com/sourcebot-dev/sourcebot/pull/1271)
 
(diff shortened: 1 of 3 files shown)


msukkari:
Superseded by #1272 (combined setup-wizard PR).

PR #1278, opened 2026-06-04
feat(web): support for anthropic prompt caching and UI cached token display

jsourcebot:
support for anthropic prompt caching and UI cached token display

## Summary by CodeRabbit

* **New Features**
  * Prompt caching for Ask Sourcebot (Enterprise) to reduce token usage — configurable and enabled by default.
  * Message details now show cached-token breakdown (count and percentage) with a tooltip explaining how many input tokens were served from the prompt cache and overall token accounting.

### DIFF
--- a/packages/web/src/ee/features/chat/agent.ts
+++ b/packages/web/src/ee/features/chat/agent.ts
@@ -197,6 +197,8 @@ export const createMessageStream = async ({
                     totalTokens: (priorMetadata?.totalTokens ?? 0) + (totalUsage.totalTokens ?? 0),
                     totalInputTokens: (priorMetadata?.totalInputTokens ?? 0) + (totalUsage.inputTokens ?? 0),
                     totalOutputTokens: (priorMetadata?.totalOutputTokens ?? 0) + (totalUsage.outputTokens ?? 0),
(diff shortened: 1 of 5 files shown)


brendan-kellam:
@jsourcebot could you include a screenshot of what the UX change looks like?

packages/web/src/ee/features/chat/agent.ts
```diff
@@ -343,11 +345,42 @@ const createAgentStream = async ({
...
+            return message;
+        }
+
+        return {
+            ...message,
+            providerOptions: {
+                ...message.providerOptions,
+                anthropic: {
+                    ...message.providerOptions?.anthropic,
+                    cacheControl: { type: 'ephemeral' },
```
brendan-kellam, line 373: what's the difference between passing the `cacheControl` option here vs., passing it as a argument in `providerOptions` we pass to streamText?

jsourcebot: cacheControl is more like a marker/break point that says "I want to cache up until this point", so don't think it logically makes sense to be passed to stream text, you really want to mark the last message you care about with this cacheControl to communicate what exactly you want cached

brendan-kellam (dismissed):

brendan-kellam (dismissed):

PR #1285, opened 2026-06-05
chore: upgrade postcss to ^8.5.15 to address CVE-2026-41305

brendan-kellam:
Fixes SOU-995

Upgrades `postcss` to `^8.5.15` to address CVE-2026-41305 (XSS via unescaped `</style>` in CSS stringify output). Refreshed the lockfile so transitive postcss instances resolve to 8.5.15. Since `next@16.2.6` hard-pins postcss at `8.4.31`, a qualified resolution keyed to that exact range redirects it to `^8.5.10`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Upgraded `protobufjs` to `^7.6.2`. [#1281](https://github.com/sourcebot-dev/sourcebot/pull/1281)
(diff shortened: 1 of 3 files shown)


brendan-kellam:
Superseded by #1288, which suppresses CVE-2026-41305 via `.trivyignore` instead of upgrading postcss.

postcss is a build-time-only dependency (no runtime stringify of untrusted CSS), so the CVE isn't exploitable here. We're opting to ignore it rather than force postcss past `next@16.2.6`'s pin of 8.4.31 with a `resolutions` override. Closing in favor of #1288.

PR #1301, opened 2026-06-11
fix(query-language): don't treat dash before a non-prefix colon word as negation

devteamaegis:
## What's broken

A search query that negates a bare word containing a colon crashes the parser, and `parseQuerySyntaxIntoIR` (strict mode) turns that into a user-facing `FAILED_TO_PARSE_QUERY` 400 — the whole search fails. Examples: `-foo:bar`, `-http://example.com`, `-time:12`. Note `foo:bar` without the dash parses fine; only the negated form breaks.

```
"-foo:bar"            => SyntaxError: No parse at 1
"-http://example.com" => SyntaxError: No parse at 1
"-file:bar"           => OK   (real prefix negation still works)
```

## Why it happens

The grammar only allows `NegateExpr { negate (PrefixExpr | ParenExpr) }`, but `negateToken` emitted the `negate` token whenever the following word contained *any* colon, not when it actually started with a known prefix keyword. So `-foo:bar` emitted `negate` and then left a bare word with no PrefixExpr to attach to → no parse.

## Fix

Only emit `negate` when the dash is immediately followed by a known prefix keyword, using the existing `startsWithPrefixAt` helper. Bare words with colons fall through to `wordToken`.

## Test

Added cases to `negation.txt`: `-http://example.com`, `-time:12`, and `repo:x -foo:bar` now parse as Terms; the existing `-file:...` prefix-negation cases still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Query parsing now correctly handles dash-prefixed terms containing colons, including URLs, without incorrectly interpreting them as negated prefixes.

* **Tests**
  * Added coverage for colon-containing terms and mixed prefix-and-term expressions to ensure accurate parsing.

---

> [!NOTE]
> **Low Risk**
> Localized query-language tokenizer change with regression tests; real prefix negation paths unchanged.
> 
> **Overview**
> Fixes **search parse failures** when a query uses a leading dash on a bare term that contains a colon (e.g. `-foo:bar`, `-http://example.com`). Those inputs previously triggered a strict-parser `SyntaxError` and `FAILED_TO_PARSE_QUERY` instead of being treated as a single search term.
> 
> **Tokenizer change:** `negateToken` no longer treats “dash + any word with a colon” as negation. It only emits `negate` when the dash is followed by a **known prefix keyword** (via `startsWithPrefixAt`), matching real negation like `-file:`. Colon-bearing bare words are left to `wordToken` as one `Term`.
> 
> **Tests & changelog:** New `negation.txt` cases cover URLs, unknown `time:` keys, and mixed `repo:x -foo:bar`; changelog entry added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95e79027549ff9b730cc43bcd2a6b32c7f542073. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


brendan-kellam:
@claude add a brief Changelog entry

brendan-kellam (approved):

PR #1348, opened 2026-06-18
feat(web): record service ping history and add usage report download

brendan-kellam:
## Overview

Offline / air-gapped deployments can't report usage to Lighthouse automatically. This change records each service ping locally so those deployments can download their usage history and send it to us out-of-band.

## Changes

- **DB**: New `ServicePingEvent` table (`id`, `payload` JSON, `createdAt`) + migration.
- **Recording**: `syncWithLighthouse` now persists each ping (via `recordServicePingInDB`) before sending it to Lighthouse, so the history is captured even when the outbound ping fails (as it always will on air-gapped instances). The **activation code is stripped** before persisting, and recording is best-effort (wrapped in try/catch + logging) so a DB error never blocks the real ping.
- **Export UI**: A "Download usage report" button on the offline license settings card (`Settings → License`), backed by an `OWNER`-gated server action that returns the recorded events. Downloads as `<date>-usage-history.json`.
- **Copy**: The offline license card explains that the instance doesn't report usage automatically and that usage must be manually sent to `ar@sourcebot.dev`.

image

## Notes

- The export button only renders for offline licenses, but recording runs on all deployments (one tiny row/day, negligible).
- No retention/pruning yet.

## Release Notes

* **New Features**
  * Offline license settings page now includes a "Download usage report" button to export usage data as JSON files.
  * Service ping history is now recorded locally for offline deployments to support usage tracking.

* **Refactor**
  * Improved license validation helper functions for better code organization.

### DIFF
--- a/packages/web/src/features/billing/servicePing.ts
+++ b/packages/web/src/features/billing/servicePing.ts
@@ -176,9 +176,7 @@ const inferDeploymentType = (): string => {
 };
 
 const recordServicePingInDB = async (payload: ServicePingRequest) => {
-    // Strip the activation code before persisting. This history is meant to be
-    // exported and sent back to us by offline deployments, so it should not
-    // contain the instance's secret activation code.
+    // Strip the activation code before persisting.
     const { activationCode: _activationCode, ...sanitizedPayload } = payload;
 
     try {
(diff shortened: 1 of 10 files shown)


packages/web/src/features/billing/servicePing.ts
```diff
@@ -83,6 +83,8 @@ export const syncWithLighthouse = async (orgId: number) => {
         ...(activationCode && { activationCode }),
     };
 
+    await recordServicePingInDB(payload);
+
     const response = await client.ping(payload);
```
msukkari, line 99: I wonder if we can skip this all together in the case where we're using an offline license?

msukkari (dismissed):

PR #1351, opened 2026-06-18
fix(gitlab): split large group sync into bounded subgroup fetches

### LINKED ISSUE #1139: [bug] Unable to sync Gitlab when there are too many repositories in a group
### Describe the bug

For example: https://gitlab.com/redhat/centos-stream/rpms
The sync timeouts.

### To reproduce

```
      gitlab-com = {
        type = "gitlab";
        token = {
          env = "GITLAB_COM_REDHAT_TOKEN";
        };
        groups = [
          "redhat/centos-stream/rpms"
          "redhat/centos-stream/tests"
        ];
      };
```

Sourcebot version 4.16

### Additional information

_No response_

RitwijParmar:
Fixes #1139

## Summary
- replace GitLab group sync's single `includeSubgroups: true` project query with recursive subgroup traversal
- fetch direct project and subgroup pages separately with bounded `perPage: 100` requests
- de-duplicate projects by GitLab project id while walking the group tree

This avoids the expensive GitLab server-side query that can time out for large namespaces such as `redhat/centos-stream/*`, while preserving recursive subgroup coverage.

## Verification
- yarn workspace @sourcebot/backend test gitlab.test.ts
- yarn workspace @sourcebot/backend build
- yarn workspace @sourcebot/backend test

## Release Notes

* **Bug Fixes**
  * Improved GitLab group synchronization for large namespaces by using bounded, offset-based pagination and explicitly traversing group subtrees.
  * Projects are collected in a consistent order and deduplicated, reducing the risk of timeouts during sync.

* **Tests**
  * Added coverage for recursive group/subgroup traversal and pagination termination, including non-advancing next-page scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgraded `tar` to `^7.5.16`. [#1338](https://github.com/sourcebot-dev/sourcebot/pull/1338)
 - Upgraded `esbuild` to `^0.28.1`. [#1342](https://github.com/sourcebot-dev/sourcebot/pull/1342)
 - Enabled Next.js version skew protection to fix "Failed to load chunk" errors during rolling deploys. [#1346](https://github.com/sourcebot-dev/sourcebot/pull/1346)
+- Split GitLab group sync into bounded direct-project and subgroup page fetches to avoid timeouts on large namespaces. [#1351](https://github.com/sourcebot-dev/sourcebot/pull/1351)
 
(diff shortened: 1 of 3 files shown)


RitwijParmar:
The queue index issue is real. I would also keep an eye on memory here because the breadth-first walk retains every subgroup id and project until the whole tree finishes. On a large GitLab namespace that can turn the timeout fix into a heap-growth problem. A next step I would take is to track queue depth and pages fetched and add a wide-tree fixture that asserts the traversal stays bounded. That gives operators something useful to act on when a sync is too large. Worth keeping in this PR or splitting out?

PR #1353, opened 2026-06-19
feat: support richer token cost tracking for ask

jsourcebot:
Screenshot 2026-06-18 at 5 43 17 PM

Adds tracking of token costs per step, also adds estimates of tool call token usage. This information is embedded in the chat history. Tool call token usages are estimates because a single step can run multiple tool calls and there is no mechanism to discern which part of the input token cost came from which tool call.

## Summary by CodeRabbit

* **New Features**
  * Per-step and per-tool token cost tracking in chat history
  * Token usage badges displaying estimated output tokens for tool calls
  * Step-indexed token usage information in chat details panel

* **Refactor**
  * Reorganized thinking steps rendering structure to support token attribution

* **Tests**
  * Added token estimation utility test suite


packages/web/src/ee/features/chat/tokenEstimation.ts
```diff
@@ -0,0 +1,63 @@
+import { ToolResultOutput } from "@ai-sdk/provider-utils";
```
brendan-kellam, line 1: should this be in /ee/features?

brendan-kellam (dismissed):

packages/web/src/ee/features/chat/tokenEstimation.ts
jsourcebot: Moved

PR #1366, opened 2026-06-23
Improve token caching.

jsourcebot:
Improved Ask Sourcebot prompt caching by splitting static and dynamic prompt sections and advancing cache breakpoints after every agent step instead of only after each message

## Summary by CodeRabbit

* **New Features**
  * Enhanced Enterprise “Ask Sourcebot” prompt caching with provider-aware strategies, separating byte-stable static vs dynamic prompt content.
  * Added server controls to enable caching, set static TTL, and optionally detect cache misses/breakpoints.
  * Cache breakpoints now advance after each agent step (not just per message).
* **Bug Fixes**
  * Stabilized prompt/tool byte layout by making MCP tool/server and repository ordering deterministic.
* **Tests**
  * Expanded coverage for cache markers, TTL behavior, no-op providers, and static prompt byte identity across repo selections.
* **Chores**
  * Updated changelog to document the improved caching behavior.

### DIFF
--- a/packages/shared/src/env.server.ts
+++ b/packages/shared/src/env.server.ts
@@ -311,13 +311,11 @@ const options = {
 
         SOURCEBOT_CHAT_MAX_STEP_COUNT: numberSchema.default(100),
         SOURCEBOT_CHAT_PROMPT_CACHING_ENABLED: booleanSchema.default('true'),
-        // Phased-rollout lever for the static front cache breakpoint (and the
-        // MCP activation-resilience breakpoint). Set to 'false' to fall back to
-        // the single moving tail marker. Only takes effect when prompt caching
-        // is enabled. See packages/web/src/ee/features/chat/promptCaching.ts.
+        // Phased-rollout lever for the static checkpoint. Set to 'false' to fall
+        // back to the single moving tail marker. Only takes effect when prompt
+        // caching is enabled.
         SOURCEBOT_CHAT_PROMPT_CACHE_STATIC_PREFIX_ENABLED: booleanSchema.default('true'),
-        // TTL for the static front block (and MCP activation-resilience block).
-        // The moving tail marker always uses the 5m default.
+        // TTL for the static block. The moving tail marker always uses the 5m default.
         SOURCEBOT_CHAT_PROMPT_CACHE_STATIC_TTL: z.enum(['5m', '1h']).default('5m'),
         // Observability: when enabled, logs a warning on unexpected prompt-cache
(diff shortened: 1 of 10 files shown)


packages/shared/src/env.server.ts
```diff
@@ -311,6 +311,16 @@ const options = {
 
         SOURCEBOT_CHAT_MAX_STEP_COUNT: numberSchema.default(100),
         SOURCEBOT_CHAT_PROMPT_CACHING_ENABLED: booleanSchema.default('true'),
+        // Phased-rollout lever for the static checkpoint. Set to 'false' to fall
```
brendan-kellam: nit: can we use /** **/ comments s.t., we get inline JSDoc rendering when hovering over these symbols in the ide?

packages/shared/src/env.server.ts
```diff
@@ -311,6 +311,16 @@ const options = {
...
+        // back to the single moving tail marker. Only takes effect when prompt
+        // caching is enabled.
+        SOURCEBOT_CHAT_PROMPT_CACHE_STATIC_PREFIX_ENABLED: booleanSchema.default('true'),
```
brendan-kellam: Is there advantage of making this configurable?

packages/web/src/ee/features/chat/promptCaching.ts
```diff
@@ -0,0 +1,184 @@
...
+import { LanguageModelProvider } from "@/features/chat/types";
+
+// @note: Prompt-cache plumbing for the Ask agent — the single place that knows
+// about provider-specific cache-control shapes. The agent loop stays
+// provider-agnostic by asking the strategy for a `providerOptions` blob at each
+// breakpoint and merging whatever it gets back (possibly `undefined`).
+
+const logger = createLogger('prompt-caching');
+
+export type CacheTtl = '5m' | '1h';
```
brendan-kellam, line 13: can the TTL only be 5m or 1h?

brendan-kellam (dismissed):

packages/shared/src/env.server.ts
jsourcebot: Mainly a safe guard in case some issue makes the static portion not cache properly, you can disable it so you stop paying the extra cost for no benefit. But it's maybe overly defensive, leaning towards removing it and keeping the env var to a minimal required set.

packages/web/src/ee/features/chat/promptCaching.ts
jsourcebot: Yes, at least for anthropic

PR #1372, opened 2026-06-26
refactor(web): resolve language model capabilities from models.dev

whoisthey:
Lays the groundwork for chat file attachments by teaching Sourcebot which inputs a configured language model can accept, resolved automatically from the [models.dev](https://models.dev) catalog (the same catalog that already backs context-window resolution).

## What this adds

* **Automatic capability resolution.** `resolveModelCapabilities` (`packages/web/src/features/chat/modelCapabilities.server.ts`) looks a model's `modalities.input` up in the models.dev catalog and partitions it into two buckets:
  * `inputModalities` (`text` | `image` | `audio` | `video`) — perceptual channels the model encodes natively.
  * `supportedDocumentTypes` (`pdf`) — rich compound container formats providers decompose server-side. models.dev folds `pdf` into `modalities.input`; we split it back out, since a document is not a modality.
* **Shared catalog module.** The catalog fetch/cache plumbing (TTL, negative cache, stale-while-revalidate, in-flight dedupe, provider-id overrides) lives in `modelsDevCatalog.server.ts` and backs both context-window and capability resolution off a single cached fetch.
* Both resolved fields are exposed on the client-safe `LanguageModelInfo` (via `getConfiguredLanguageModelsInfo` and the MCP ask path).

This is pure capability plumbing. It adds no attachment UI (future work).

## Behavior

* **Catalogued models** light up their real capabilities automatically, with no `config.json` changes.
* **Uncatalogued models** (e.g. `openai-compatible` / self-hosted endpoints) fall back to text-only with no document support: the model stays fully usable for normal chat, and richer attachments stay gated off until support can be positively confirmed.


jsourcebot (dismissed):

packages/web/src/features/chat/types.ts
```diff
@@ -209,10 +209,15 @@ type _AssertAllProviders = LanguageModelProvider extends typeof languageModelPro
...
 void _assertAllProviders;
 
+export type InputModality = 'text' | 'image' | 'audio' | 'video';
+export type DocumentType = 'pdf';
+
 export const languageModelInfoSchema = z.object({
     provider: z.enum(languageModelProviders).describe("The model provider (e.g., 'anthropic', 'openai')"),
     model: z.string().describe("The model ID"),
     displayName: z.string().optional().describe("Optional display name for the model"),
+    inputModalities: z.array(z.enum(['text', 'image', 'audio', 'video'])).default(['text']).describe("The input modalities the model can accept (images, audio, video, text). Single-medium attachments are gated by these. Defaults to text-only."),
```
brendan-kellam: nit: could we have the type derive from the enum s.t., these are guaranteed to remain in sync?

brendan-kellam (dismissed):

packages/web/src/features/chat/types.ts
brendan-kellam: @whoisthey double check if this is going to break back-compat with existing chats

whoisthey: Nice, changed to be derived from type

whoisthey: Backwards compat green from spot check, and from code it stands because of the defaults

PR #1397, opened 2026-06-29
feat(web): pin+flag Ask citations to a resolved commit SHA

whoisthey:
Pins each Ask file citation to the commit SHA it was sourced at (`git rev-parse` for `read_file`; the repo's indexed commit for `grep`/`glob`/symbol search), and fetches the evidence panel + copied links at that SHA. Keeps citation content and line ranges aligned with the code as it was when answered, rather than drifting against HEAD. `commitSha` **is optional,** so older chats fall back to the symbolic ref.

Groundwork for a follow-up freshness hint. `get_diff`/`list_tree` left unpinned for now.

Should be basically indistinguishable UI/UX difference, just now that any code referenced in the evidence panel should be as-the-agent-saw-it. In the event that the revision was lost in git history, it will fallback to one retry fetch from effectively `latest`, which is the existing behaviour.

## Additional

Adds a per-citation freshness hint: `/api/source/freshness` compares a pinned commit to the repo's current default-branch tip (file-level blob comparison), and the evidence panel shows a badge when the cited file has `changed` / `removed` / `pinned_unavailable`, with a "View latest" link. Also resolves a citation to a pinned source when multiple tools sourced the same file in one turn.

Scope: panel-only (no inline-chip dots or answer rollup yet); `list_tree`/`get_diff` remain unpinned.

(removed the "view latest" button from these two, but here's what the badges look like for the other two statuses (not included in the screenrecording)

Screenshot 2026-06-29 at 2 19 13 PM
Screenshot 2026-06-29 at 2 19 27 PM

https://github.com/user-attachments/assets/8eeca34b-f101-4fb6-bef6-9ca108447d93

## Summary by CodeRabbit

* **New Features**
  * File citations are now pinned to the commit they were generated from, keeping the displayed line ranges aligned over time.
  * Added file citation staleness hints plus a freshness indicator in the referenced file view (including “view latest” when available).
  * Search and developer tools (symbols, references, grep, glob, and file reads) now surface commit-specific source details for more precise referencing.
* **Bug Fixes**
  * When a pinned file ref is invalid, the app falls back to an alternate revision to keep content rendering.

### DIFF
--- a/packages/web/src/features/git/schemas.ts
+++ b/packages/web/src/features/git/schemas.ts
@@ -35,6 +35,8 @@ export const fileSourceResponseSchema = z.object({
     repoExternalWebUrl: z.string().optional(),
     webUrl: z.string(),
     externalWebUrl: z.string().optional(),
+    // The concrete commit SHA that `ref` resolved to. Undefined if unresolvable.
+    commitSha: z.string().optional(),
 });
 
 export const getDiffRequestSchema = z.object({
--- a/packages/web/src/features/git/schemas.ts
+++ b/packages/web/src/features/git/schemas.ts
@@ -39,6 +39,24 @@ export const fileSourceResponseSchema = z.object({
     commitSha: z.string().optional(),
 });
 
+export const fileFreshnessRequestSchema = z.object({
+    repo: z.string(),
+    path: z.string(),
+    // The pinned commit SHA a citation was sourced at.
+    sinceSha: z.string(),
+});
+
+export const fileFreshnessResponseSchema = z.object({
+    // `fresh`: the cited file is byte-identical at the current tip.
+    // `changed`: the file's content differs since the pinned commit.
+    // `removed`: the file no longer exists at the current tip.
+    // `pinned_unavailable`: the pinned commit is no longer in the repo (e.g.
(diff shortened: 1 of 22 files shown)


packages/web/src/features/git/schemas.ts
```diff
@@ -35,6 +35,8 @@ export const fileSourceResponseSchema = z.object({
     repoExternalWebUrl: z.string().optional(),
     webUrl: z.string(),
     externalWebUrl: z.string().optional(),
+    // The concrete commit SHA that `ref` resolved to. Undefined if unresolvable.
+    commitSha: z.string().optional(),
```
brendan-kellam, line 39: since this is is exposed in our public API, we'll want to update the openapi spec with `yarn openapi:generate` (we should automate this).

brendan-kellam: @claude open a PR to add a github action that runs `openapi:generate` on PRs, creating a commit if there is a modification

brendan-kellam: https://github.com/sourcebot-dev/sourcebot/pull/1398

whoisthey: Updated spec on this branch for now 👍

PR #1410, opened 2026-07-02
Jminnetian/askskills

jsourcebot:
Added Ask Sourcebot skills, letting users create, import, share, sync, and auto-invoke reusable chat instructions across personal and workspace scopes.

## Summary by CodeRabbit

* **New Features**
  * Added Ask Sourcebot Skills for creating, importing, sharing, syncing, and invoking reusable chat instructions.
  * Added personal and shared skill management, including slash-command suggestions, workspace controls, adoption, publishing, and repository imports.
  * Added repository sync status, previews, updates, and source-access visibility.
  * Added automatic skill invocation in chat with loaded-skill details and file references.
* **Documentation**
  * Added comprehensive Skills documentation covering creation, sharing, syncing, permissions, and usage.
* **Bug Fixes**
  * Improved unsaved-change navigation protection, tooltip behavior, and chat interaction states.

---

> [!NOTE]
> **Medium Risk**
> Large new EE surface area touches chat prompts, authorization (personal/shared/synced repo visibility), and persistence; incorrect scope checks or command materialization could leak instructions or change model behavior broadly.
> 
> **Overview**
> Introduces **Ask Sourcebot skills**: reusable markdown instructions with slash commands, scoped as **personal** or **shared** workspace catalog entries, with **adoptions** and owner **auto-enroll** controls.
> 
> **Data & API:** New `AgentSkill` / `AgentSkillAdoption` Prisma models and migration; shared query helpers in `@sourcebot/db`. EE server actions and `GET /api/ee/skills/sourceStatus` support CRUD, publishing, repo/file import, and sync via `sourceBlobSha` (also exposed on search `blobSha` in OpenAPI).
> 
> **Chat behavior:** Skill commands load into `ChatBox` on landing and thread pages; user messages carry `data-command` parts that **`materializeCommandMessageTexts`** expands before streaming. The agent gets a sanitized **`<agent_skills>`** catalog, a **`load_skill`** tool, and uses materialized instruction text in the model context; chat details render loaded skills. Skill usage analytics events are added alongside MCP.
> 
> **UI:** **Settings → Skills** (entitlement-gated), workspace **Ask Sourcebot** shared-skills manager, connectors menu link, and docs (`skills.mdx`). Smaller fixes: app-level **`NavigationGuardProvider`**, tooltip portal defaults, search mode tooltip gating, and OpenAPI `blobSha` for synced imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74c48b5c2160f586f03bccf7303f6a8060cdada0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/db/prisma/schema.prisma
```diff
@@ -741,6 +746,93 @@ model ChatAccess {
...
+  instructions String
+
+  // The user who authored the skill. This is audit metadata, separate from
+  // the namespace that owns the slash-command slug.
+  createdBy   User   @relation("AgentSkillCreatedBy", fields: [createdById], references: [id], onDelete: Cascade)
+  createdById String
+
+  updatedBy   User?   @relation("AgentSkillUpdatedBy", fields: [updatedById], references: [id], onDelete: SetNull)
+  updatedById String?
+
```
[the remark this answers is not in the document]
jsourcebot: As of current this was actually intentional, but it's unclear if it should be this way or if a shared skill truly becomes an org resource (but then it's weird that the creator is the sole person who can then remove everybodies access to it).

packages/db/prisma/migrations/20260706000100_add_agent_skills/migration.sql
brendan-kellam, line 1: could we compress into a single migration?

brendan-kellam:
image

nit: looks like the "x" button is not centered vertically

brendan-kellam:
image


Not something we need to do now, but it would be cool if the icon was clickable and it revealed a icon/emoji dropdown, similar to Linear:

image

packages/web/src/lib/posthogEvents.ts
```diff
@@ -241,6 +241,22 @@ export type PosthogEventMap = {
         durationMs: number,
         failureReason?: string,
     },
+    // Fired when an agent skill is invoked, either automatically by the model
+    // (activationMethod: 'auto', via the load_skill tool) or manually by the
+    // user (activationMethod: 'manual', via a slash command). Multi-source, so
+    // no wa_ prefix; source identifies the origin.
+    ask_skill_invoked: {
```
msukkari, line 258: are we sending posthog events for other instances (ex. create skill, edit skill, share skill, etc) @jsourcebot ? we should add events for every interaction

jsourcebot: No, I can add more extensive posthog logging though

msukkari: sg thanks, we should try to add posthog events for every major action

packages/web/src/app/(app)/settings/components/settingsContentFrame.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+
+import { usePathname } from "next/navigation";
+
+// Routes that opt out of the centered settings column and render edge-to-edge
+// (e.g. the full-bleed skill editor). Everything else stays in the constrained,
+// centered column shared by the rest of settings.
+const FULL_BLEED_ROUTE_PATTERNS = [
```
brendan-kellam: could we include the members page as part of this pattern?

brendan-kellam: maybe we want to standardize around the existing `SettingsContainer`?

brendan-kellam:
@jsourcebot could you include UX screenshots in the PR description?

packages/web/src/ee/features/chat/tools/loadSkillTool.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Single-phase skill loader, modelled on Claude Code's `Skill` tool. The model
+ * discovers skills from the `<agent_skills>` catalog in the system prompt, then
+ * calls this tool with the skill's id to load its (auth-checked) instructions as
+ * the tool result. Unlike MCP tools, a
+ * skill is an instruction blob rather than a callable tool with its own schema,
+ * so it does not need to be promoted into `activeTools` — one call returns the
+ * instructions directly. This keeps the tool set static across steps, so the
+ * prompt cache is never invalidated by skill loading.
+ */
+export const createLoadSkillTool = ({
```
brendan-kellam, line 42: Thoughts on: for the purposes of consistency in the codebase, would it make sense for this tool (as well as `tool_request_activation` tool) be defined using the `ToolDefinition` interface and be located next to the other tools in `/features/tools`?

packages/web/src/ee/features/chat/skills/skillCommands.server.ts
brendan-kellam, line 1: this fils is marked `.server.ts` - is this accurate?

packages/web/src/components/ui/tooltip.tsx
```diff
@@ -15,15 +15,17 @@ const TooltipContent = React.forwardRef<
...
-  <TooltipPrimitive.Content
-    ref={ref}
-    sideOffset={sideOffset}
-    className={cn(
-      "z-50 overflow-hidden rounded-md border bg-popover px-3 py-1.5 text-sm text-popover-foreground shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 origin-[--radix-tooltip-content-transform-origin]",
-      className
-    )}
-    {...props}
-  />
+  <TooltipPrimitive.Portal>
```
brendan-kellam, line 18: Heads up: iirc I had done this before, but it broke certain Tooltips in the app, so you might want to double check this isn't going to break anything. Additionally, I know there are some tooltips (e.g., upgradeBadge.tsx) that are already wrapping the Tooltip in a portal.

packages/web/src/ee/features/chat/skills/components/skillsPage.tsx
brendan-kellam, line 1: nit: from a maintainability & readability pov, consider breaking this file up a bit

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1626 @@
...
+            listSharedAgentSkillCommandsForContext({ prisma, userId: user.id, orgId: org.id }),
+        ]);
+
+        return [
+            ...personalCommands,
+            ...sharedCommands,
+        ];
+    }));
+
+export const createSharedAgentSkill = async (
```
brendan-kellam, line 1386: I see that you're using this server action as part of a form submission, which in that case makes sense in terms of validating against a schema. In other parts of the codebase, we are using `useForm`, which is a hook that allows you to validate a form schema client-side. After it's been validated client-side, we just call a server action with direct parameters.

We may want to consider doing that instead, because it's kind of the pattern that we've been using in the codebase thus far.

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1626 @@
...
+                emitSkillEvent('ask_skill_created', eventBase, {
+                    success: false,
+                    failureReason: ErrorCode.UNEXPECTED_ERROR,
+                });
+                throw error;
+            }
+        }));
+};
+
+export const updateSharedAgentSkill = async (
```
brendan-kellam, line 1445: same comment as above

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1626 @@
...
+            actorRelationship: getSharedSkillActorRelationship(existingSkill, user.id, role),
+        }, { success: true });
+        return { success: true };
+    }));
+
+export const setSharedSkillFlag = async (
+    input: SharedSkillFlagInput,
+    analytics?: SkillAnalyticsContext,
+): Promise<SharedAgentSkillManagementItem | ServiceError> => {
+    const parsed = sharedSkillFlagInputSchema.safeParse(input);
```
brendan-kellam, line 1565: same comment as above

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1626 @@
...
+            return askError;
+        }
+
+        const sharedSkill = await prisma.agentSkill.findFirst({
+            where: {
+                id: skillId,
+                ...sharedAgentSkillAuthScope(org.id),
+                enabled: true,
+            },
+            select: {
```
brendan-kellam, line 1211: do we need to do this select? Applies to other parts of this file.

packages/web/src/ee/features/chat/skills/components/skillInstructionsEditor.tsx
```diff
@@ -0,0 +1,399 @@
...
+            )}
+        </>
+    );
+};
+
+const DefaultElement = (props: RenderElementProps) => (
+    <p {...props.attributes} className="m-0">{props.children}</p>
+);
+
+const FileMentionComponent = ({
```
brendan-kellam: nit: this feels like it should be shared with whatever we are doing in the Chatbox

brendan-kellam (dismissed):

packages/web/src/ee/features/chat/tools/loadSkillTool.ts
jsourcebot: Hmm, using the interface is probably a good idea. I dont know about changing the location.. these arent tools that our mcp server exposes, they are enterprise only tools used by the Ask agent. Might opt to leave them where they are.

That being said, at a glance, the interface also isn't a great match for these tools, I would have to generalize/modify the ToolDefinition interface.

Edit: I went ahead and refactored the tools into their own files, but I did not use the ToolDefinition interface, they just arent the same kind of tools and dont follow the same structure, dont want to change the existing mcp server tools too much atm, can revisit in the future if more "meta" tools like the two we already have get made.

packages/web/src/ee/features/chat/skills/skillCommands.server.ts
jsourcebot: Yea, it's only used by the server, the goal is to pull the skills available to the current user which is based on auth/db look up. I will rename it though to be more clear that these are purely for the skills commands. More general "available to all" commands in the future would not be fetched from the server in this way.

packages/web/src/components/ui/tooltip.tsx
jsourcebot: I removed the portal from upgradeBadge tooltip as well as another. It seems composing them like what happened here was safe, but there was no reason to keep it that way since it was just tech debt after my tooltip change,

packages/web/src/ee/features/chat/skills/actions.ts
jsourcebot: Hmm, I feel like server side validation is the right thing to be doing. I dont mind also validating the schema client side if that follows the code base pattern, but I am not confident about just trusting whatever the client sends after. Maybe that is overly cautious? But in general with the desire to support a public deployment and keep it stable, I think also validating schemas on the server should be the pattern we adopt long term.

brendan-kellam: I guess my point is that server actions already enforce a contract for the expected input types, so validating them with a schema seems redundant. API routes are a different story and should always have their request bodies and query parameters validated by a schema.

packages/web/src/ee/features/chat/skills/actions.ts
jsourcebot: The selects mainly help with explicitly stating which fields we are interested in for subsequent operations, and also improving retrieval performance (latter depends on how many fields are skipped). I don't have a particular preference but I am leaning on the side of leaving this as is.

packages/web/src/ee/features/chat/skills/actions.ts
jsourcebot: Not sure I follow/agree? I am assuming the expected input types is properly validated for "types" but doesnt do check for things like empty strings, slug format, max lengths? I think that is the benefit coming from the server side validation.

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1728 @@
...
+                ...personalAgentSkillAuthScope(user.id, org.id),
+            },
+        });
+
+        if (!skill) {
+            return skillNotFound();
+        }
+
+        return toAgentSkillListItem(skill);
+    }));
```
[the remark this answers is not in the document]
jsourcebot: Intended for now. This is more in line with the direction of "synced skills do not require permissions" that we likely want to move towards.

packages/web/src/app/(app)/settings/components/settingsContentFrame.tsx
jsourcebot: I went ahead and reused SettingsContainer as you suggested

packages/web/src/ee/features/chat/skills/actions.ts
```diff
@@ -0,0 +1,1728 @@
...
+        };
+        try {
+            const personalSkill = await prisma.$transaction(async (tx) => {
+                const createdSkill = await tx.agentSkill.create({
+                    data: {
+                        ...personalAgentSkillScope(user.id, org.id),
+                        slug: sharedSkill.slug,
+                        name: sharedSkill.name,
+                        description: sharedSkill.description,
+                        instructions: sharedSkill.instructions,
```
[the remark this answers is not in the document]
jsourcebot: Interesting catch, while the UI does not actually support it, it is indeed possible to copy a shared skills into a personal copy if you manually request it. This and many other things all just point to the fact that we should not really be imposing repo visibility rules on skills, the overhead/complexity is quite high and it also makes life difficult for users. I will not address this right now, it's clear that it's best to just remove the repo permission gates on skills in a follow up PR.

packages/web/src/ee/features/chat/skills/commandResolution.ts
```diff
@@ -0,0 +1,396 @@
...
+        if (!repo || !path) {
+            continue;
+        }
+
+        const source: FileSource = {
+            type: "file",
+            repo,
+            path,
+            name: path.split("/").pop() || path,
+            revision: DEFAULT_SKILL_FILE_REFERENCE_REVISION,
```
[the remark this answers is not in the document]
jsourcebot: This is fine, we will not be snapshotting the version at which the file reference was when the skill mentioning it was created. Using the HEAD reference is desirable/acceptable

packages/web/src/ee/features/chat/skills/actions.ts
brendan-kellam: > do check for things like empty strings, slug format, max lengths? I think that is the benefit coming from the server side validation.

yep this is a good callout, and in favor in what you are doing here. Maybe in the future, this is the model we want to move to for all APIs that sit at the server/client boundary (both REST apis & server actions), where the expectation being that they define a input & output schema. I was [playing around with this idea in Lighthouse](https://github.com/sourcebot-dev/lighthouse/blob/4d74555c81a5a7849bcadc45927973cae0bded65/lambda/utils.ts#L147), and found it has worked pretty well.

PR #1415, opened 2026-07-02
fix(backend): decode percent-encoded repo names for generic Git URLs

### LINKED ISSUE #1384: Generic Git URL configs keep percent-encoded repo names
## Problem
Generic Git connection configs that point directly at an HTTP(S) remote derive repo names from the raw URL pathname. If the remote path contains encoded characters, such as `%20`, Sourcebot stores names like `github.com/test/Project%20Name` instead of the decoded repo name.

## Reproduction
Configure a generic Git URL such as `https://github.com/test/Project%20Name%20With%20Spaces.git` and compile the connection config.

## Expected
The direct URL path should match the existing file-based generic Git behavior and derive `github.com/test/Project Name With Spaces`.

## Actual
The direct URL path keeps `%20` in `name`, `displayName`, and zoekt metadata.

## Impact
The same remote can produce inconsistent Sourcebot/zoekt repo identifiers depending on whether it is configured as a direct URL or discovered from a local repository origin.

rachit367:
Fixes #1384

## Problem

A generic Git connection pointing directly at an HTTP(S) remote whose path contains encoded characters (e.g. `https://github.com/test/Project%20Name%20With%20Spaces.git`) stores the repo name with the encoding intact — `github.com/test/Project%20Name%20With%20Spaces` — in `name`, `displayName`, and the zoekt metadata.

`compileGenericGitHostConfig_url` derives the name from `remoteUrl.pathname` without decoding, while the file-origin path (`compileGenericGitHostConfig_file`) already calls `decodeURIComponent`. So the **same remote** produces different Sourcebot/zoekt identifiers depending on whether it's configured as a direct URL or discovered from a local repository origin.

(Re: @brendan-kellam's question on the issue — that inconsistency across the two code paths is the concrete impact.)

## Fix

Decode `remoteUrl.pathname` in `compileGenericGitHostConfig_url` before building the name, mirroring the existing file-origin behavior.

## Tests

Added a `compileGenericGitHostConfig_url` case asserting a `%20`-encoded URL yields `github.com/test/Project Name With Spaces` in `name`, `displayName`, and `zoekt.name` — matching the existing `_file` test. `tsc --noEmit` is clean for the backend package.

(Note: the repo's generic-git `_url` tests assert forward-slash-joined names and only pass on POSIX CI, since `path.join` uses the platform separator; they fail identically on Windows before this change. The decode logic itself is platform-independent.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Maintained the sidebar scroll position when navigating between chats instead of resetting to the top. [#1411](https://github.com/sourcebot-dev/sourcebot/pull/1411)
 - Upgraded `nodemailer` to `^9.0.1`. [#1356](https://github.com/sourcebot-dev/sourcebot/pull/1356)
 - Upgraded `@opentelemetry/core` to `^2.8.0`. [#1413](https://github.com/sourcebot-dev/sourcebot/pull/1413)
+- Decoded percent-encoded characters in repo names derived from generic Git URLs, so direct URLs match the file-based behavior. [#1415](https://github.com/sourcebot-dev/sourcebot/pull/1415)
 
 ## [5.0.4] - 2026-06-18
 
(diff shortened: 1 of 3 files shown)


rachit367:
Closing in favor of #1389, which predates this (opened 2026-06-29) and fixes the same issue more completely — it wraps the decode in a helper that preserves malformed percent-escapes (e.g. `Project%GGName`) instead of letting `decodeURIComponent` throw during config compilation, which my version here doesn't handle. Deferring to @DivyamTalwar's PR. Thanks!

PR #1529, opened 2026-08-01
fix(web): scope file-search recents to (repo, revision) in the browse dialog

### LINKED ISSUE #1387: File search recents are shared across browse revisions
## Problem
The browse file-search dialog stores recently opened files under a key scoped only by repository name. When a user switches branches or revisions, recents from another revision can still appear.

## Reproduction
Open file search for a repo on one revision, select a file, then switch to another branch or revision in the same repo and open file search again.

## Expected
Recently opened files should be scoped to the revision context that can resolve those paths.

## Actual
The dialog can show repo-level recents from a different revision. Selecting one navigates with the current revision, which can point to a path that does not exist there.

## Impact
Users browsing multiple branches can get stale file suggestions and confusing dead-end navigation.

Harsh23Kashyap:
## Summary

The browse file-search dialog (`mod+p`) stored recently opened files in `localStorage` under a key scoped only by `repoName`. A user who opened a file on `main`, switched to `feature/foo`, and re-opened the dialog would see recents from `main` — paths that may not exist on the new revision. Selecting one navigated with the new revision and landed on a 404.

Fix: scope the `localStorage` key to the `(repoName, revisionName)` tuple, so a branch switch in the same repo yields a fresh recents list scoped to the new revision.

Fixes #1387.

## What changes

`packages/web/src/app/(app)/browse/components/fileSearchCommandDialog.tsx:38` — the `useLocalStorage` key changes from `` `recentlyOpenedFiles-${repoName}` `` to `` `recentlyOpenedFiles-${repoName}-${revisionName ?? 'HEAD'}` ``. The `'HEAD'` default matches the file-fetch fallback on the next line, so the recents key and the file list key agree on the "no revision in URL" case.

The `setRecentlyOpened` callback does not need to change — `useLocalStorage` from `usehooks-ts` writes to whatever key the current key string resolves to. Switching revisions automatically re-renders the dialog with the new key, which starts empty (or with whatever recents the user previously stored under that exact `(repo, revision)` tuple).

Old keys (`` `recentlyOpenedFiles-<repo>` ``) become orphaned entries in `localStorage` and are ignored. No migration is required — the worst case is the user loses the recents accumulated under the old key, which is the bug being fixed.

## Files

- `packages/web/src/app/(app)/browse/components/fileSearchCommandDialog.tsx` — the key-derivation change (one line, plus a 4-line comment explaining the rationale).
- `packages/web/src/app/(app)/browse/components/fileSearchCommandDialog.test.tsx` — 3 new vitest cases.
- `CHANGELOG.md` — one-sentence entry under `[Unreleased] → Fixed`.

## Why this is in scope

The component already reads `revisionName` from `useBrowseParams()` on the line above and uses it to fetch the file list (`getFiles({ repoName, revisionName: revisionName ?? 'HEAD' })`), so the data was already revision-scoped. The `localStorage` key was the only thing that wasn't. The fix is a single key-derivation change that brings the persistence layer in line with the data layer that's already there.

## Test coverage

3 vitest cases in `fileSearchCommandDialog.test.tsx`:

- **Scopes the recents localStorage key to the `(repo, revision)` tuple.** Asserts the key is `recentlyOpenedFiles-github.com/foo/bar-main` for `(github.com/foo/bar, main)`. The regression assertion for the bug.
- **Uses the `HEAD` fallback for the "no revision in URL" case.** The file-fetch on the next line uses `revisionName ?? 'HEAD'` as the default; the recents key needs to agree so the user sees a consistent recents list for the "default branch" view. Asserts the key is `recentlyOpenedFiles-github.com/foo/bar-HEAD` when `revisionName` is `undefined`.
- **Produces a different key for a different revision in the same repo.** The bug was: switching from `main` to `feature/foo` showed recents from `main`. After the fix, the keys differ, so the recents are naturally scoped. Asserts `main` and `feature/foo` produce different keys for the same repo.

The test stubs `useLocalStorage` from `usehooks-ts` to capture the key the component passes (without this, we'd be asserting on real `localStorage`, which jsdom does provide but is per-test mutable state that's harder to reason about). The other hooks the dialog uses (`useBrowseParams`, `useBrowseState`, `useQuery`, etc.) are stubbed so the test only exercises the key derivation.

3/3 tests pass; 7 pre-existing OpenTelemetry-setup failures in `ee/askmcp/...` and `ee/permissionSyncStatus/...` are unchanged by this PR (they fail to *load* due to the OTel SDK version mismatch, before any of my code runs).

## Backward compatibility

Pure behaviour change. Old keys' data is left in `localStorage` and ignored, so a user upgrading from a previous version simply starts with an empty recents list on their next branch switch. No data loss on the server side.

## Risks

Minimal. The default `'HEAD'` matches the file-fetch default on the next line, so the recents key and the file list key agree on the "no revision in URL" case. If a future change splits `'HEAD'` and `undefined` semantics, both sides will need to be updated in lockstep — but that's already true for the file fetch.

## Future work

- A "clear recents" UI button next to the recents list. Not in scope for this PR; the bug was about cross-revision leakage, not about retention control.
- Migrating recents to the server (Prisma `userRecentFile` table scoped by `(userId, repoId, revisionName)`). Punted: a `localStorage` key is the lowest-cost fix and matches the existing pattern. If a user clears their browser data they lose their recents, but they also lose their session, so this is fine.

---

> [!NOTE]
> **Low Risk**
> Client-only localStorage key change in the browse file-search dialog; no auth, API, or server impact beyond users starting fresh recents per revision.
> 
> **Overview**
> Fixes browse **mod+p** file search showing **recently opened** paths from another branch after a revision switch (issue #1387).
> 
> **`FileSearchCommandDialog`** now persists recents under a `localStorage` key derived from **`[repoName, revisionName ?? 'HEAD']`** via `recentlyOpenedFiles::${JSON.stringify(...)}`, instead of repo name only. That aligns recents with the same revision default used for `getFiles`, and avoids ambiguous keys that a simple `repo-revision` string could collide on. Prior repo-only keys are left unused (no migration).
> 
> Adds **`fileSearchCommandDialog.test.tsx`** (Vitest) to lock the key format, `HEAD` when revision is missing, per-revision separation, and collision safety. **CHANGELOG** documents the fix under `[Unreleased]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b0eb9aeeb8ef4f71c007c12faa8ad3c91c7d572. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Browse file-search suggestions and recently opened files are now scoped to the selected repository and revision.
  * Suggestions no longer carry over when switching branches or revisions.
  * Searches without a specified revision consistently use `HEAD`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Upgraded `brace-expansion` to `^1.1.17`/`^2.1.3`/`^5.0.8`. [#1527](https://github.com/sourcebot-dev/sourcebot/pull/1527)
-- The browse file-search dialog's recently-opened files are now scoped to the `(repo, revision)` tuple, so switching branches in the same repo no longer carries over suggestions from a different revision that may not exist on the current one. [#1387](https://github.com/sourcebot-dev/sourcebot/issues/1387)
+- The browse file-search dialog's recently-opened files are now scoped to the `(repo, revision)` tuple, so switching branches in the same repo no longer carries over suggestions from a different revision that may not exist on the current one. [#1529](https://github.com/sourcebot-dev/sourcebot/pull/1529)
 
 ## [5.1.5] - 2026-07-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Upgraded `brace-expansion` to `^1.1.17`/`^2.1.3`/`^5.0.8`. [#1527](https://github.com/sourcebot-dev/sourcebot/pull/1527)
+- The browse file-search dialog's recently-opened files are now scoped to the `(repo, revision)` tuple, so switching branches in the same repo no longer carries over suggestions from a different revision that may not exist on the current one. [#1387](https://github.com/sourcebot-dev/sourcebot/issues/1387)
 
 ## [5.1.5] - 2026-07-31
 
--- a/packages/web/src/app/(app)/browse/components/fileSearchCommandDialog.test.tsx
+++ b/packages/web/src/app/(app)/browse/components/fileSearchCommandDialog.test.tsx
@@ -1,0 +1,122 @@
+import { afterEach, describe, expect, test, vi } from 'vitest';
+import { render } from '@testing-library/react';
+
+// Stub `useLocalStorage` so the test can capture the key the component
+// passes and assert on it. Without this, we'd be asserting on real
+// localStorage (which jsdom does provide, but the key changes per
+// revision and the test only wants to verify the key format).
+const capturedKeys: string[] = [];
+vi.mock('usehooks-ts', () => ({
+    useLocalStorage: <T,>(key: string, initial: T) => {
+        capturedKeys.push(key);
+        return [initial, vi.fn()] as [T, (v: T) => void];
+    },
+}));
+
+// Make the params mock swappable per-test. We use a single mutable
+// holder so each `it` can set its own (repoName, revisionName) tuple
+// before rendering.
+const mockParams: { repoName: string; revisionName: string | undefined } = {
+    repoName: 'github.com/foo/bar',
+    revisionName: 'main',
+};
+vi.mock('@/app/(app)/browse/hooks/useBrowseParams', () => ({
+    useBrowseParams: () => ({ repoName: mockParams.repoName, revisionName: mockParams.revisionName }),
+}));
+vi.mock('@/app/(app)/browse/hooks/useBrowseState', () => ({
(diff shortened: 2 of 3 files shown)


Harsh23Kashyap:
Both CodeRabbit inline comments addressed in [4b0eb9ae](https://github.com/sourcebot-dev/sourcebot/pull/1529/commits/4b0eb9ae):

- **CHANGELOG link** — fixed; was the issue URL (`#1387`), now points at the PR (`#1529`).
- **Collision-free key encoding** — the previous `<repoName>-<revisionName>` template collides for tuples like `(`foo-bar`, `baz`)` vs `(`foo`, `bar-baz`)` since both components can contain `-`. Switched to `recentlyOpenedFiles::${JSON.stringify([repoName, revisionName ?? 'HEAD'])}`. JSON.stringify of a 2-element array is a unique representation of the tuple, so the key is collision-free regardless of which characters appear in the components. The `::` prefix and `HEAD` default are preserved.

Added a 4th test case that asserts the two ambiguous tuples above produce different keys.

4/4 tests pass.

